PR #2, opened 2021-04-15
Generate object node classes from manifest

eslavich:
This PR generates object node class definitions from the rad datamodels-1.0 manifest.

I also added some testing infrastructure and a simple test to make sure the generated classes have the necessaries.


src/roman_datamodels/stnode.py
```diff
@@ -252,30 +261,66 @@ def get_schema(self):
...
+    _OBJECT_NODE_CLASSES.append(cls)
+    globals()[class_name] = cls
+
+_OBJECT_NODE_CLASSES_BY_TAG = {c._tag: c for c in _OBJECT_NODE_CLASSES}
+
+
 class TaggedObjectNodeConverter(Converter):
     """
-    This class is intended to be subclassed for specific tags
+    Converter for all subclasses of TaggedObjectNode.
```
perrygreenfield, line 309: "Converter for all subclasses of TaggedObjectNode that don't require special handling"?

src/roman_datamodels/stnode.py
```diff
@@ -252,30 +261,66 @@ def get_schema(self):
...
 
     @property
     def tag(self):
         return self._tag
 
 
+_DATAMODELS_MANIFEST_PATH = importlib_resources.files(rad.resources) / "manifests" / "datamodels-1.0.yaml"
+_DATAMODELS_MANIFEST = yaml.safe_load(_DATAMODELS_MANIFEST_PATH.read_bytes())
+_OBJECT_NODE_CLASS_NAME_OVERRIDES = {
+    "tag:stsci.edu:datamodels/roman/reference_files/flat-1.0.0": "FlatRef",
```
perrygreenfield, line 275: There are going to be a lot of these. Should there be more generic handling of reference file tags?

src/roman_datamodels/stnode.py
perrygreenfield: An example of a case that does use special handling (e.g., for wfi_mode, where a property for filters is added) would be good since the following uses the same converter class for all nodes. That example would be essentially done as before, and added to extensions as before?

perrygreenfield (commented):
Fantastic. My only comments are about being explicit on how to handle cases that aren't generic object nodes. Since I know little about tox, ci.yml, and publish-to-pypi.yml files, I can't offer any useful review of those, but am grateful for your providing them.

src/roman_datamodels/stnode.py
eslavich: That's a good idea, shall I just add `Ref` to the ends of their names?

src/roman_datamodels/stnode.py
eslavich: Oh right, we're going to need a list of exceptions or something.  I'll see what I can cook up.

src/roman_datamodels/stnode.py
perrygreenfield: Yes.

eslavich: This is done

src/roman_datamodels/stnode.py
eslavich: I added a custom `WfiMode` that uses properties to map `optical_element` to one of `filter` or `grating`.  This kind of implementation allows us to still use the same converter.

perrygreenfield (approved):
LGTM

PR #6, opened 2021-04-28
Changes made to deal with CRDS needs and a move of where test utilities are located

perrygreenfield:
Map the start_time to date and time values for CRDS, move the test utilities so they can be accessed by other packages.


src/roman_datamodels/datamodels.py
```diff
@@ -242,10 +242,17 @@ def get_crds_parameters(self):
...
         """
-        return {
+        crds_header = {
             key: val for key, val in self.to_flat_dict(include_arrays=False).items()
             if isinstance(val, (str, int, float, complex, bool))
         }
+        # We need to add two distinct time and date entries derived from the start time
+        # since that is what CRDS expects.
+        time = crds_header['roman.meta.observation.start_time']
+        date, time = str(time).split('T')
```
eslavich, line 252: Since we're depending on the T, I'd call `time.isot` here instead of `str(time)`, in case someone decides to use one of the other astropy time formats:

```python
In [38]: Time.now()
Out[38]: <Time object: scale='utc' format='datetime' value=2021-04-29 13:29:34.763737>

In [39]: str(Time.now())
Out[39]: '2021-04-29 13:29:37.987641'

In [40]: Time.now().isot
Out[40]: '2021-04-29T13:29:45.209'
```

src/roman_datamodels/datamodels.py
```diff
@@ -242,10 +242,17 @@ def get_crds_parameters(self):
...
         dict
         """
-        return {
+        crds_header = {
             key: val for key, val in self.to_flat_dict(include_arrays=False).items()
             if isinstance(val, (str, int, float, complex, bool))
         }
+        # We need to add two distinct time and date entries derived from the start time
+        # since that is what CRDS expects.
+        time = crds_header['roman.meta.observation.start_time']
```
eslavich, line 251: This value won't be included in `crds_header`, since we're only selecting instances of simple built-in types.

src/roman_datamodels/datamodels.py
```diff
@@ -242,10 +242,17 @@ def get_crds_parameters(self):
         -------
         dict
         """
-        return {
+        crds_header = {
             key: val for key, val in self.to_flat_dict(include_arrays=False).items()
```
eslavich, line 246: Does `to_flat_dict` return keys that start with "roman."?  If not then we'll need to modify the key here.

tests/test_models.py
```diff
@@ -8,7 +8,7 @@
 from roman_datamodels import stnode
 from roman_datamodels.extensions import DATAMODEL_EXTENSIONS
 
-import util
+from roman_datamodels import test_utils as util
```
eslavich, line 11: There is a `roman_datamodels.testing` module that this could go in, that's where I tucked away the factories (modeled after `numpy.testing`)

src/roman_datamodels/datamodels.py
perrygreenfield: I thought that was what you were requesting due to other parts of CRDS looking at the raw asdf, where roman is the top attribute. I can remove roman, but is that what you want? All the current selection criteria include it.

tests/test_models.py
perrygreenfield: You mean that test_utils.py should go in the testing package? Or the that test_models.py should also go there?

src/roman_datamodels/datamodels.py
eslavich: Right, that's what I was confirming, CRDS does want the keys to start with `roman.`.  I wasn't sure if `to_flat_dict` would do that since in `DataModel.__init__` we're pulling out the roman node and assigning it to `_instance`.  But I see now that the `to_flat_dict` implementation is adding "roman." back in:

```python
        if include_arrays:
            return dict(('roman.' + key, convert_val(val)) for (key, val) in self.items())
        else:
            return dict(('roman.' + key, convert_val(val)) for (key, val) in self.items()
                        if not isinstance(val, np.ndarray))
                        ```

tests/test_models.py
eslavich: The former.  It seems tidy to put all the testing tools into that one `roman_datamodels.testing` namespace (`roman_datamodels.testing.util` in this case).  It's not a big deal though.

src/roman_datamodels/datamodels.py
eslavich: Did a commit get lost here?  This conversation and the one about `crds_header` are marked resolved, but I don't see any code changes that address them?

perrygreenfield: No, but some evil force interposed between my fingers and the editor save command thereby holding hostage the relevant changes. I have assuaged the evil force with appropriate sacrifices. It should be in the PR now.

eslavich (approved):
LGTM

PR #11, opened 2021-06-03
Update models and factories for latest rad changes

eslavich:
This incorporates the most recent wave of changes that were merged in rad.


src/roman_datamodels/testing/factories.py
```diff
@@ -13,6 +13,7 @@
 import numpy as np
 
 from .. import stnode
+from .. import tables
```
PaulHuwe, line 16: Maybe we should rename tables.py to something a bit more descriptive or specific (just because "tables" is used in so many of our packages)? Maybe something like definitiion_tables.py? Not a hard issue, just a suggestion.

src/roman_datamodels/testing/factories.py
```diff
@@ -143,6 +144,23 @@ def _random_array_uint32(size=(4096, 4096), min=None, max=None):
     return np.random.randint(min, high=max, size=size, dtype=np.uint32)
 
 
+def _random_dq_def(size=10, dq_size=32):
+    table = np.zeros((size,), dtype=tables.DQ_DEF_DTYPE)
```
PaulHuwe, line 148: Rename to dq_table? Again, for ease of reading

src/roman_datamodels/testing/factories.py
```diff
@@ -409,6 +425,102 @@ def create_flat_ref(**kwargs):
...
+    **kwargs
+        Additional or overridden attributes.
+
+    Returns
+    -------
+    roman_datamodels.stnode.DarkRef
+    """
+    raw = {
+        "data": _random_array_float32((4096, 4096, 1)),
+        "dq": _random_array_uint16(),
```
PaulHuwe, line 444: Why isn't this 4k x 4k?

src/roman_datamodels/testing/factories.py
```diff
@@ -409,6 +425,102 @@ def create_flat_ref(**kwargs):
...
+    **kwargs
+        Additional or overridden attributes.
+
+    Returns
+    -------
+    roman_datamodels.stnode.MaskRef
+    """
+    raw = {
+        "meta": create_ref_meta(reftype="MASK"),
+        "dq": _random_array_uint16(),
```
PaulHuwe, line 493: Same question here about 4k x 4k

src/roman_datamodels/testing/factories.py
```diff
@@ -457,6 +569,30 @@ def create_guidestar(**kwargs):
...
+    -------
+    dict
+    """
+
+    return {
+        "calibration_software_version": _random_string("Version ", 120),
+        "crds_context_used": "roman_{:04d}.pmap".format(_random_positive_int(9999)),
+        "crds_software_version": _random_software_version(),
+        "filename": _random_string("Filename ", 120),
+        "file_date": _random_astropy_time(),
```
PaulHuwe, line 587: Not an issue now I suspect, but aren't random dates problematic around CRDS? (If this ever tests against CRDS, that is..)

src/roman_datamodels/testing/factories.py
eslavich: It is, due to the `_random_array_uint16` default for the size argument.

src/roman_datamodels/testing/factories.py
eslavich: Sure, I'll call this dq_def since that seems to be the name it's know by

src/roman_datamodels/testing/factories.py
eslavich: Yeah, they would be.  I expect that the random values will be refined over time as they prove to be problematic (ideally with associated changes to the schemas).

src/roman_datamodels/testing/factories.py
eslavich: What I have in mind for this module (eventually) is a method for each Roman table that creates an empty table object in the correct format.  We can test them against the schemas in this repo and then people who create reference files don't have to mess with dtypes.  Does that sound useful?

PaulHuwe: Yes it does.

perrygreenfield (approved):
LGTM

src/roman_datamodels/testing/factories.py
eslavich: Per conversation on Slack changing this module name to "table_definitions"

PaulHuwe (approved):
TGTM

PR #13, opened 2021-06-14
Make 'tag' parameter name consistent with parent class' method

mcara:


mcara:
Here is the link to the parent class: https://github.com/asdf-format/asdf/blob/887c5bec20541e6585acb128b0a17a089254c2f5/asdf/extension/_converter.py#L83-L120

Both `to_yaml_tree(self, obj, tag, ctx)` and `from_yaml_tree(self, node, tag, ctx)` have `tag : str` which is a single string.

eslavich (approved):
LGTM

PR #15, opened 2021-06-26
Rcal 173 ramp wfi image rdm support

PaulHuwe:
Added support and factories for the following models:
- Ramp
- Rampfitoutput
- WFIImgPhotom


eslavich (approved):
LGTM

PR #20, opened 2021-08-02
RCAL-181 Added ScienceRawModel, factory, and model maker utility.

PaulHuwe:
This ticket adds RDM support of ScienceRawModel needed for the romancal DQ_init step.

This is pointing to the code in PR RAD-64.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.4.0 (unreleased)
 ==================
 
-- Added support for ScienceRawModel. [#20]
+- Added support for ScienceRawModel. Removed basic from ref_common in testing/utils. [#20]
 
 - Added support for dq_init step in cal_step. [#18]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 0.4.0 (unreleased)
 ==================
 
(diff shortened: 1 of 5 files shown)


nden:
@PaulHuwe Can you describe how this is going to be used or point to the code in dq_init which uses it.

ddavis-stsci (approved):
Looks good. I think that factories.py will also need a change. 
In line 1037 it seems to be using a 4-d array and not a 3-d

PaulHuwe:
1037 was already changed from 4D to 3D.

perrygreenfield (approved):
Other than Paul's comment it looks good to me.

nden:
Just to make sure I understand this correctly - this model is to be used with raw files, so the data array needs to be 3D.
When the file is read into a RampModel a dimension will be added to the data array as expected by the pipeline.
If so why is the ramp schema defining the data array as ndim=3?

perrygreenfield:
If you want that done automatically, then the converter needs to deal with that. The default machinery doesn't handle that. I think the optical element case is an example of how it is handled in the data model converters. I'll have to look.

nden:
For now the cal code does this. So I'm going to merge this PR and make a release.

PR #24, opened 2021-08-11
RCAL 183 Added gain maker utility and tests.

PaulHuwe:
Create maker utility function for gain
Create tests for gain utility function


ddavis-stsci (approved):
LGTM

PR #26, opened 2021-08-13
RCAL 162 Added dark current model maker and tests.

PaulHuwe:
Create maker utility function
Create tests for maker utility function


ddavis-stsci (approved):
LGTM

PR #30, opened 2021-08-24
Prepare for 0.5.1 release

nden:


PaulHuwe (approved):
Why 0.5.1 instead of 0.6.0 where it would match RAD?

nden:
Any reason they need to match? I see roman_datamodels changing more frequently in the future.

PaulHuwe:
Ah, that makes sense. Would we want the major number of RAD & RDM to match? (I.e. RDM 0.6.X would line up with RAD 0.6.0... this may help with confusion over compatible RDM & RAD versions, since they are closely intertwined?)

Issue #32, opened 2021-09-03
Datamodels should allow creation of new attributes


perrygreenfield:
While the node machinery allows this, the top level datamodel class does not. It should.

WilliamJamieson:
@perrygreenfield was this competed by #33?

PR #33, opened 2021-09-03
added ability to add attributes to datamodels

perrygreenfield:
Added the ability to add attributes to datamodels (using the same mechanism as for TaggedObjectNodes, i.e., by using dictionary indices).

Closes [#32]

### DIFF
--- a/src/roman_datamodels/datamodels.py
+++ b/src/roman_datamodels/datamodels.py
@@ -184,7 +184,11 @@ class DataModel:
         if key.startswith('_'):
             raise ValueError(
                 'May not specify attributes/keys that start with _')
-        self._instance._data[key] = value
+        if hasattr(self._instance, key):
+            print('XXXXXX')
+            setattr(self._instance, key, value)
+        else:
+            self._instance._data[key] = value
 
     def to_flat_dict(self, include_arrays=True):
         """
--- a/src/roman_datamodels/datamodels.py
+++ b/src/roman_datamodels/datamodels.py
@@ -185,7 +185,6 @@ class DataModel:
             raise ValueError(
                 'May not specify attributes/keys that start with _')
         if hasattr(self._instance, key):
-            print('XXXXXX')
             setattr(self._instance, key, value)
         else:
             self._instance._data[key] = value
--- a/src/roman_datamodels/datamodels.py
+++ b/src/roman_datamodels/datamodels.py
@@ -19,6 +19,7 @@ from . import stnode
 from . import validate
 from . extensions import DATAMODEL_EXTENSIONS
 
+
 class DataModel:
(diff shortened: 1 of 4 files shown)


tests/test_models.py
```diff
@@ -201,18 +206,30 @@ def test_opening_mask_ref(tmp_path):
...
 def test_opening_readnoise_ref(tmp_path):
     # First make test reference file
     file_path = tmp_path / 'testreadnoise.asdf'
     utils.mk_readnoise(filepath=file_path)
     readnoise = datamodels.open(file_path)
     assert readnoise.meta.instrument.optical_element == 'F158'
     assert isinstance(readnoise, datamodels.ReadnoiseRefModel)
+
+
+def test_add_model_attribute(tmp_path):
```
eslavich, line 227: Do we need to be able to save these new attributes to the .asdf file, and read them back again?  That's worth a test, if so.

PaulHuwe (commented):
LGTM

src/roman_datamodels/datamodels.py
```diff
@@ -179,6 +180,12 @@ def __setattr__(self, attr, value):
     def __getattr__(self, attr):
         return getattr(self._instance, attr)
 
+    def __setitem__(self, key, value):
```
eslavich, line 183: What happens if the user sets a key from the schema using the bracket operator?  Does this bypass validation?

Also, can the new attribute be set with dot notation after it's been created?

tests/test_models.py
perrygreenfield: Sure, though I don't see why that won't work. But I'll add a test.

src/roman_datamodels/datamodels.py
perrygreenfield: I'll check the first, probably it does, so I should see if the attribute exists first. Yes, it should be modifiable by dot notation, but also a good thing to test.

src/roman_datamodels/datamodels.py
```diff
@@ -179,6 +180,16 @@ def __setattr__(self, attr, value):
...
+        if key.startswith('_'):
+            raise ValueError(
+                'May not specify attributes/keys that start with _')
+        if hasattr(self._instance, key):
+            print('XXXXXX')
```
eslavich, line 188: Sir, I caught this print statement nosing around the methods.  Do you want me to send it to the brig?

eslavich (approved):
One rogue line of code, but otherwise LGTM

src/roman_datamodels/datamodels.py
perrygreenfield: Damn, I'm so imperfect. No, I'll delete it.

Issue #34, opened 2021-09-03
Datamodels constructors should raise an exception if the top level node of an asdf file has the wrong type.


perrygreenfield:
The current code does not check to see, when using a data model constructor, that the asdf file provided has a top node that has a type consistent with the data model class. The implementation to provide this should allow for specific cases where they are mismatched for conversion functions to be used, but this is expected to be infrequent.

Currently there exists a stub method to do this, but it currently does nothing.

nden:
Fixed in #35

PR #35, opened 2021-09-13
Validate if the tag type in the file does not match the class used to open the file

perrygreenfield:
The changes here address issue #34. Now if this happens, a ValueError is raised.

Note that this does not use the normal validation machinery but instead explicitly compares the class types. Is there a good reason to do it otherwise?


ddavis-stsci (approved):
LGTM

PR #37, opened 2021-09-29
Added support for Saturation reference files.

PaulHuwe:
Implement datamodel support for saturation reference files. 

Created maker utility for the data model.

Created tests for the maker utility.


PaulHuwe:
The test failures are expected due to RDM being behind RAD in terms of updates at the moment.

ddavis-stsci (approved):
Looks Good.

PR #38, opened 2021-09-29
RCAL-223 Update RDM for 2D Ramp Pedestal Array and Misc

PaulHuwe:
Update Ramp Pedestal Array to 2D. Fixed reference model casting in test_models.

Test failures are expected as RDM is behind RAD in updates.


ddavis-stsci (approved):
LG

PR #39, opened 2021-10-05
RCAL-206 Implement Nonlinearity RDM Support

PaulHuwe:
Implemented support and tests for linearity reference model. 
Corrected dimension order in factories. 
Added primary array definition to MaskRefModel.

NOTE: Test fails are expected as RDM is behind RAD in developments.


ddavis-stsci (approved):
Looks good.

PR #42, opened 2021-10-07
RCAL-225 Update RDM Tests for Mode and Optical_Element in Reference Files

PaulHuwe:
Updated tests and makers for exposure and optical_element requirements in reference files.


PaulHuwe:
Failed tests are expected until RDM catches up to RAD.

PR #53, opened 2021-11-10
Add support for cal_logs

eslavich:
This adds support for cal_logs-1.0.0, which is an array of log strings from the calibration pipeline that produced the file.


perrygreenfield (approved):
LGTM

PR #63, opened 2022-02-03
Fix bug with asdf 2.9.x due to change in private variable name

eslavich:
The `TagDefinition._schema_uri` variable changed to `TagDefinition._schema_uris`.  This updates roman_datamodels to use the public property instead of the unreliable private variable.


perrygreenfield (approved):
Looks good to me.

PR #64, opened 2022-02-11
Changes to models testing functions to bring them in sync with schemas

nden:
The changes in schemas include
- photometry changes - spacetelescope/rad#114
- making photometry an optional property - spacetelescope/rad#115
- adding `p_keywords` - spacetelescope/rad#105
- adding `level0_compression` attribute - spacetelescope/rad#104
- adding `gw_fgs_mode` to `guide_star` - spacetelescope/rad#103

This PR also updates versioons of some dependencies in setup.cfg and points to rad/main

### DIFF
--- a/src/roman_datamodels/testing/utils.py
+++ b/src/roman_datamodels/testing/utils.py
@@ -436,10 +436,13 @@ def mk_level1_science_raw(shape=None, filepath=None):
 
     Parameters
     ----------
-    shape
-        (optional) Shape of arrays in the model.
+    shape : tuple, int
+        (optional) (z, y, x) Shape of data array. This includes a four-pixel
+        border representing the reference pixels. Default is (8, 4096, 4096)
+        (8 integrations, 4088 x 4088 represent the science pixels, with the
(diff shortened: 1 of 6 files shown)


src/roman_datamodels/testing/utils.py
```diff
@@ -397,7 +399,7 @@ def mk_basic_meta():
     return meta
 
 
-def mk_common_meta():
+def mk_common_meta(photometry=True):
```
ddavis-stsci, line 402: Should this be photmetry=False as the default if we are removing it as a required element? and set it to true once we want it?

ddavis-stsci (approved):
Besides my comment below (which is really about style) LGTM.

cshanahan1 (approved):

src/roman_datamodels/testing/utils.py
nden: removed from `common` now

nden:
This PR includes the commits from #65.
It also points to the rad/main until a rad release is made.

PR #69, opened 2022-02-28
Fixed import issue

WilliamJamieson:
`asdf.fits_embed` is not available from an `import asdf`. This was only working by accident before. This PR resolves this bug.


WilliamJamieson:
This is an alternate fix to asdf-format/asdf#1097.

ddavis-stsci (approved):
LGTM

PR #71, opened 2022-03-16
RCAL-321: Remove exptype and p_keywords from Distortion DataModel Utilities

PaulHuwe:
Removed exptype and p_keyword from Distortion maker utility and factory.


ddavis-stsci (approved):
LGTM

PR #72, opened 2022-03-31
RCAL-325: Updating Photom DataModels in RDM

PaulHuwe:
Updating photom maker utilities and tests.

### DIFF
--- a/src/roman_datamodels/testing/factories.py
+++ b/src/roman_datamodels/testing/factories.py
@@ -665,49 +665,49 @@ def create_wfi_img_photom_ref(**kwargs):
     """
     raw_dict = {
         "F062":
-            {"photmjsr": (10 * np.random.random() * u.MJ / u.sr),
-             "uncertainty": np.random.random() * u.MJ / u.sr,
-             "pixelareasr": .2 * u.sr},
+            {"photmjsr": (1.0e-15  * np.random.random() * u.MJ / u.sr),
+             "uncertainty": (1.0e-16  * np.random.random() * u.MJ / u.sr),
+             "pixelareasr": 1.0e-13 * u.sr},
         "F087":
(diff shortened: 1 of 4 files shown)


src/roman_datamodels/testing/factories.py
```diff
@@ -663,18 +664,50 @@ def create_wfi_img_photom_ref(**kwargs):
     roman_datamodels.stnode.WfiImgPhotomRef
     """
     raw_dict = {
+        "F062":
+            {"photmjsr": (10 * np.random.random() * u.MJ / u.sr),
```
ddavis-stsci, line 668: Can this be something like,
"photmjsr": (1.0e-15 * np.random.random() * u.MJ / u.sr
and
"uncertainty": 1.0e-15 * np.random.random() * u.MJ / u.sr
To get the magnitude more or less correct.

src/roman_datamodels/testing/factories.py
```diff
@@ -663,18 +664,50 @@ def create_wfi_img_photom_ref(**kwargs):
...
+             "uncertainty": np.random.random() * u.MJ / u.sr,
+             "pixelareasr": .2 * u.sr},
```
ddavis-stsci, line 670: This is a pixel area of 0.2 sr? I think that something like
"pixelareasr":  1.0e-13 * u.sr}
might be closer to the final values.

ddavis-stsci (approved):
The one thing I'd update are the default values. If I'm reading this correctly the default values would be 1-10 for photmjsr, 0-1 for uncertainty and the pixel area is 0.2 steradians. It might be better to have more realistic values. I've put some suggestions below.

src/roman_datamodels/testing/factories.py
PaulHuwe: Updated - also updated the maker utility to match.

PR #73, opened 2022-04-08
Rcal-326: Correct Photom Units

PaulHuwe:
Corrected photom units to megajanskies.


src/roman_datamodels/testing/factories.py
```diff
@@ -1038,6 +1044,49 @@ def create_ramp_fit_output(**kwargs):
...
+    raw["meta"]["gw_end_time"] = _random_astropy_time()
+    raw['meta']['gw_frame_readout_time'] = _random_float()
+    raw['meta']['pedestal_resultant_exp_time'] = _random_float()
+    raw['meta']['signal_resultant_exp_time'] = _random_float()
+    raw['meta']['gw_acq_number'] = _random_int()
+    raw['meta']['gw_mode'] = 'WIM-ACQ'
+    raw['meta']['gw_window_xstart'] = _random_float()
+    raw['meta']['gw_window_ystart'] = _random_float()
+    raw['meta']['gw_window_xstop'] = _random_float()
+    raw['meta']['gw_window_ystop'] = _random_float()
```
ddavis-stsci, line 1083: Shouldn't the stop values be something like 
gw_window_xstart +16 or 32 just so they are about the right size?

src/roman_datamodels/testing/factories.py
```diff
@@ -395,7 +396,8 @@ def create_ref_meta(**kwargs):
             "optical_element": _random_optical_element(),
         },
         "observation": {
-            "ma_table_name": "ma_table.name",
+            "ma_table_name": _random_string("MA table "),
+            "ma_table_number":  _random_positive_int(),
```
ddavis-stsci, line 400: maybe add a range here? 1-999

ddavis-stsci (approved):
Overall looks good, two minor comments

src/roman_datamodels/testing/factories.py
PaulHuwe: Changed.. also adjusted size to match.

PaulHuwe: Also adjusted the maker utility.

PR #74, opened 2022-04-11
RCAL-314: Update roman_datamodels to include ma_table_number

PaulHuwe:
Moved ma_table_name and ma_table_number from observation to exposure groups.


ddavis-stsci (approved):
LGTM

PR #77, opened 2022-04-25
Define files for test coverage

nden:


PaulHuwe (commented):
LGTM - do you intend to add a comment for this in the changelog?

nden:
@PaulHuwe I don't think it needs a change log since it's not a user facing change.

PR #79, opened 2022-04-26
Copy assertion for mode equality from asdf-astropy

nden:
This PR copies `asdf-astropy.testing.helpers.assert_model_equal` to `roman_datamodels.testing.assertions`.
The testing functions in roman_datamodels are used by CRDS tests. However, CRDS tests do not use pytest and CRDS has no dependency on pytest. Importing from asdf-astropy adds pytest as a dependency of CRDS which is undesirable.

Alternatively this can be fixed in asdf-astropy in a future release because the pytest dependency is coming from an import from astropy.units which can be done within a function. However, in the interest of releasing this for build 6, the function is copied now here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+0.12.2 (2022-04-26)
+===================
+
+- Added function for model equality. [#79]
+
 0.12.1 (2022-04-26)
 ===================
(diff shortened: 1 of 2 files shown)


alphasentaurii (approved):
LGTM - thanks!

PR #82, opened 2022-06-08
RCAL-347: Update roman_datamodels util ramp error array creation

PaulHuwe:
Updated utilities and test for change in dimensionality of err variable in ramp datamodel.


ddavis-stsci (approved):
LGTM

PR #85, opened 2022-07-18
freeze action versions

zacharyburnett:
freezes the versions of actions used in the GitHub Actions workflow, to ensure that API changes don't break the unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,12 +116,12 @@ jobs:
           - python: '3.10'
             numpy: '1.20.*'
     steps:
-      - uses: actions/checkout@v3.0.2
-      - uses: actions/setup-python@v4.1.0
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         id: python
         with:
           python-version: ${{ matrix.python }}
-      - uses: actions/cache@v3.0.5
+      - uses: actions/cache@v3
         with:
           path: ${{ env.pythonLocation }}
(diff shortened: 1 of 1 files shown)


WilliamJamieson (changes requested):
I don't like the exact version pins I think we can just pin to a major version (`v3`  in all cases I think).

.github/workflows/ci.yml
```diff
@@ -116,12 +116,12 @@ jobs:
           - python: '3.10'
             numpy: '1.20.*'
     steps:
-      - uses: actions/checkout@main
-      - uses: actions/setup-python@main
+      - uses: actions/checkout@v3.0.2
```
WilliamJamieson, line 119:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/ci.yml
```diff
@@ -116,12 +116,12 @@ jobs:
...
+      - uses: actions/setup-python@v4.1.0
```
WilliamJamieson, line 120:
```suggestion
      - uses: actions/setup-python@v4
```

.github/workflows/ci.yml
```diff
@@ -116,12 +116,12 @@ jobs:
...
         id: python
         with:
           python-version: ${{ matrix.python }}
-      - uses: actions/cache@main
+      - uses: actions/cache@v3.0.5
```
WilliamJamieson, line 124:
```suggestion
      - uses: actions/cache@v3
```

WilliamJamieson (changes requested):

PR #87, opened 2022-07-27
Fix times, change W146 to F146, move 'gw_function_start/end' end and 'gw_acq_exec_stat' from guidestar to guidewindow

cshanahan1:
This PR has a few changes:

- The factory function for creating 'exposure' was setting the exposure start and end times as floats because they were never converted from 'unix' format to 'isot'. There are several other instances where this is happening (when the `_random_astropy_time` function is called), for example in the 'useafter' value, I'm not sure which ones need to be changed so I just fixed the instances in exposure for now.

- Changed filter name from W146 to F146.

- Moved 'gw_function_start', 'gw_function_end,' and 'gw_acq_exec_stat' from factory functions for guidestar to guidewindow ( see https://github.com/spacetelescope/rad/pull/154/).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@
 
 - Add support for new ``rad`` schema tags. [#86]
 
+- Fixed format of exposure times factory functions, changed filter 'W146' to 'F146'. [#87]
+
 0.12.2 (2022-04-26)
 ===================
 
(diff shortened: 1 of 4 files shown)


nden:
One of the filter names has changed W146->F146.

nden (approved):

ddavis-stsci (approved):
LGTM

PaulHuwe (approved):
LGTM

PR #91, opened 2022-08-17
pin `asdf` to newest patch version to fix dependency for newest `jsonschema` release

zacharyburnett:
related to https://github.com/asdf-format/asdf/pull/1171; when `asdf 2.12.1` is released, this PR can be merged

PR #92, opened 2022-08-17
Update asdf version to fix jsonschema issues

WilliamJamieson:
Updates asdf version to deal with issue asdf-format/asdf#1172.


WilliamJamieson:
Duplicates #91.

PR #94, opened 2022-08-19
Enable pre-commit and add autoformatting

WilliamJamieson:
Adds pre-commit and enables automatic formatting and other checks. These include:
- black
- isort
- pyupgrade
- autoflake8
- bandit
- Trailing whitespace fixes


zacharyburnett:
I use `isort --fass --fss .` personally as I think it makes the imports clearer

WilliamJamieson:
> I use `isort --fass --fss .` personally as I think it makes the imports clearer

No I don't like this at all. It mixes `from ... import` with `import ...` stuff.

PR #96, opened 2022-08-23
Fix ref_file util

WilliamJamieson:
Fixes the ref_file util to fit with spacetelescope/rad#157 changes to the schema.


ddavis-stsci (approved):
Looks good and the tests pass, see
https://plwishmaster.stsci.edu:8081/me/my-views/view/all/job/RT/job/Roman-devdeps/115/

PR #98, opened 2022-08-29
RCAL-356 Add test for `stuserdict.py` to `roman_datamodels`

zacharyburnett:
resolves [RCAL-356](https://jira.stsci.edu/browse/RCAL-356)


ddavis-stsci (approved):
Looks good.

PR #100, opened 2022-08-29
move build configuration into `pyproject.toml`

zacharyburnett:
similar to https://github.com/spacetelescope/romancal/pull/512

`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

This functionality with `setuptools` will likely be supported upon the release of Python 3.11, with its native implementation of `tomli`

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.


WilliamJamieson:
Needs a slight modification to work with pyproject.toml configured projects, `isolated_build = True` needs to be added see: https://github.com/asdf-format/asdf/blob/3e6679f310cdaabddb55761880c952d0a5199b35/tox.ini#L1-L3.

docs/conf.py
```diff
@@ -40,8 +38,8 @@ def setup(app):
 sys.path.insert(0, os.path.abspath('roman_datamodels/'))
 
 # -- General configuration ------------------------------------------------
-conf.read([os.path.join(os.path.dirname(__file__), '..', 'setup.cfg')])
-setup_cfg = dict(conf.items('metadata'))
+conf = tomli.load('../pyproject.toml')
```
WilliamJamieson, line 41:
```suggestion
with open(Path(__file__).parent.parent / "pyproject.toml", "rb") as configuration_file:
    conf = tomli.load(configuration_file)
```

`from pathlib import Path` will need to be added at the top as well

pyproject.toml
```diff
@@ -1,6 +1,93 @@
...
+    'stsci-aws-utils >= 0.1.2',
+]
+docs = [
+    'sphinx',
+    'sphinx-automodapi',
+    'sphinx-rtd-theme',
+    'stsci-rtd-theme',
+    'sphinx-astropy',
+    'sphinx-asdf',
+]
```
WilliamJamieson, line 42:
```suggestion
docs = [
    'sphinx',
    'sphinx-automodapi',
    'sphinx-rtd-theme',
    'stsci-rtd-theme',
    'sphinx-astropy',
    'sphinx-asdf',
    'tomli; python_version <"3.11"`,
]
```

nden:
Note #108

pyproject.toml
```diff
@@ -1,6 +1,94 @@
...
+authors = [{ name = 'STScI', email = 'help@stsci.edu' }]
+classifiers = [
+    'Intended Audience :: Science/Research',
+    'Topic :: Scientific/Engineering :: Astronomy',
+    'License :: OSI Approved :: BSD License',
+    'Programming Language :: Python :: 3',
+]
+dependencies = [
+    'asdf >=2.13.0',
+    'asdf-astropy >=0.2.0',
```
zacharyburnett, line 16:
```suggestion
    'asdf-astropy >=0.2.0',
    'gwcs >=0.18.1',
```
from #108

WilliamJamieson (approved):
LGTM

PR #101, opened 2022-09-07
Initial Association datamodel support.

PaulHuwe:
As the title says.. this is an initial take on a data model support for associations. This is a companion ticket to the schema ticket: 
https://github.com/spacetelescope/rad/pull/162
I am opening this in draft PR to facilitate conversation.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,8 @@ install_requires =
     psutil>=5.7.2
     numpy
     astropy>=5.0.4
-    rad>=0.13.1
+    #rad>=0.13.1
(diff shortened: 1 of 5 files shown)


WilliamJamieson:
If I understand this correctly, there is some sort of json file which gets read to create the `AssociationsModel`? If so I don't see how this is being done because right now it looks like it will try to open one using the asdf file opening mechanism. While json should be valid yaml and most valid yaml will work as an asdf file, there are some things that asdf expects in the file in order to open it which will be missing.

PaulHuwe:
The method to input via json model isn't coded yet (that is ticket RCAL-395) - there is merely a placeholder comment until that functionality is implemented.

PR #103, opened 2022-09-11
RCAL-400 Investigate Simplified ModelContainer DataModel

PaulHuwe:
This is a rough investigation of a simplified ModelContainer datamodel.

### DIFF
--- a/src/roman_datamodels/datamodels.py
+++ b/src/roman_datamodels/datamodels.py
@@ -7,13 +7,16 @@ keep consistency with the JWST data model version.
 It is to be subclassed by the various types of data model variants for products
 '''
 from pathlib import PurePath
+import builtins
 import datetime
 import sys
 import asdf
 import os
 import os.path
 import copy
+import json
 import numpy as np
+from collections.abc import Sequence
 from astropy.time import Time
 from asdf.fits_embed import AsdfInFits
 from . import stnode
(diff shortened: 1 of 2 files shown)


src/roman_datamodels/datamodels.py
```diff
@@ -300,6 +303,57 @@ class GuidewindowModel(DataModel):
     pass
 
 
+class ModelContainerModel(DataModel, Sequence):
+    # Needs a number of methods to properly handle the contents
+    def __init__(self, init=None, asn_schema= None, asn_file_path=None, model_file_path=None, iscopy=False, **kwargs):
```
nden, line 308: It looks like there's a mixture of parameters - those used to initialize the model from association table and also from some other input. Add a documentation section to clarify what the parameters are.

src/roman_datamodels/datamodels.py
```diff
@@ -300,6 +303,57 @@ class GuidewindowModel(DataModel):
...
+        #                  iscopy=False, **kwargs):
+        super().__init__(init=None, **kwargs)
+
+        self._models = []
+        self._iscopy = iscopy
+        # self.asn_exptypes = asn_exptypes
+        # self.asn_n_members = asn_n_members
+        # self.asn_table = {}
+        # self.asn_table_name = None
+        # self.asn_pool_name = None
```
nden, line 319: Aren't the above attributes needed?

src/roman_datamodels/datamodels.py
```diff
@@ -300,6 +303,57 @@ class GuidewindowModel(DataModel):
...
+
+        self._memmap = kwargs.get("memmap", False)
```
nden, line 321: Memory mapping was disabled in jwst. My recollection is this was due to difficulties with FITS. Is it necessary to disable it for ASDF files?

src/roman_datamodels/datamodels.py
```diff
@@ -300,6 +303,57 @@ class GuidewindowModel(DataModel):
...
+        if asn_file_path:
+            with builtins.open(asn_file_path, 'r') as asn_file:
+                asn_schema = json.load(asn_file)
+
+        for product in asn_schema['products']:
+            for member in product['members']:
+                if self._iscopy:
+                    member_model = open(model_file_path + member['expname'])
+                else:
+                    member_model = "null"
```
nden, line 334: Could you add a comment here why the above logic is necessary? It's not obvious to me.
Also it's worth investigating the use of `_iscopy` in jwst and whether it's needed for roman.

nden: Also this section of the code assumes the input is an association table and that asn_schema is created in the previous code block or passed as an argument in the init. However, if the input is a list of models, or both `asn_schema` and `asn_file_path` are None  this will crash.

ddavis-stsci: And it does not even seem to be working with an association,

```
>>> a2 = builtins.open('r00001-o001_image_001_asn.json')
>>> c1 = rdm.datamodels.ModelContainerModel(a2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.10/site-packages/roman_datamodels/datamodels.py", line 329, in __init__
    for product in asn_schema['products']:
TypeError: 'NoneType' object is not subscriptable
```

PaulHuwe:
Closing this DPR, as it is just a proof of concept. The code is going to be re-used as the base for the actual implementation, in RCAL-401.

PR #99, opened 2022-08-29
RCAL-357 Add a test for `filetype.py` to `roman_datamodels`

zacharyburnett:
[RCAL-357](https://jira.stsci.edu/browse/RCAL-357)

PR #104, opened 2022-09-12
properly handle string input to `filetype.check()`

zacharyburnett:
The `check` function as modified in https://github.com/spacetelescope/roman_datamodels/pull/99 only checked `os.PathLike` and `pathlib.Path`. This fixes that issue and adds a test for the case of string input.


WilliamJamieson (approved):
LGTM

PR #108, opened 2022-09-30
Add `gwcs` as a dependency

### LINKED ISSUE #107: Should we add `gwcs` as a dependency.
An issue arose where someone could not open L2 pipeline results using just `roman_datamodels` the issue turned out to be that `gwcs` was missing from the environment. Thus should `gwcs` be a dependency of `roman_datamodels`

WilliamJamieson:
As stated in #107, once cannot read L2 files using `roman_datamodels` only. This is because they include `gwcs` information. Since `roman_datamodels` provides the ASDF file read/write for support for all the data models used by Roman installing it should require all of the ASDF file read/write support libraries for the objects which are contained with the data models themselves which are written to the files.

Fixes #107

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.14.0 (unreleased)
 ===================
 
--
+- Explicitly add ``gwcs`` to the list of dependencies. [#108]
 
 0.13.0 (2022-08-23)
 ===================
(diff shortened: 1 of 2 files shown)


perrygreenfield (approved):
LGTM

nden:
Good! We should be able to open files without the pipeline being installed.

PR #109, opened 2022-10-12
RCAL 415: Add support for special roman units

WilliamJamieson:
Adds `roman_datamodels` support for spacetelescope/rad#168. This will enable `roman_datamodels` to serialize `DN` and `electron` units, as serialization of those units cannot be supported by `asdf-astropy`.

Note in order to serialize `DN` and `electron` one needs to use `roman_datamodels.units.DN` or `roman_datamodels.units.electron` respectively instead of `astropy.units.DN` or `astropy.units.electron`. The versions provided by `roman_datamodels` are mapped in `astropy` to be equivalent to their respective `astropy` units. However, they provide the necessary hooks to serialize them for Roman.


WilliamJamieson:
This PR requires the merger of spacetelescope/rad#168 in order for the `test latest commit` CI to pass.

WilliamJamieson:
As it turns out @PaulHuwe found that we actually need to serialize the composite unit `electron / s` which requires support to serialize these roman non-VOunits. I added support for that as far as possible.

The issue is when defining a composite unit with a non-VOunit for roman we must always have the roman unit as the left most unit when building it. Should I add documentation concerning these units? If so here or in `rad`?

schlafly (approved):
Looks good to me.  The tests look excellent; I should try to emulate them.

How annoying is adding additional units?  I tried for a while to think of any other non-VO units that Roman might need, but I wasn't able to come up with any.  But I could imagine we've missed something and it would be good to know how annoying it is.  Just adding to ROMAN_UNIT_SYMBOLS looks straightforward?

And just food for thought... some things are unitless but kind of have units.  e.g., you take an image in uncalibrated electrons and multiply it by the flat field with units of calibrated electrons / uncalibrated electrons and get out something with units of calibrated electrons.  I don't think we should do anything there (i.e., I'm not asking for new calibrated / uncalibrated "units"!), but I guess I'm just saying that we should make sure that the reference files, even after adding units, still contain information that explains the content of the files a bit.  Only tangentially related to this PR.

PaulHuwe (approved):
Looks good, modulo code coverage. Passed RDM unit tests, RCAL unit tests, RCAL regression tests. I was able to utilize them some on python on the side.

WilliamJamieson:
> Looks good to me.  The tests look excellent; I should try to emulate them.
> 
> 
> 
> How annoying is adding additional units?  I tried for a while to think of any other non-VO units that Roman might need, but I wasn't able to come up with any.  But I could imagine we've missed something and it would be good to know how annoying it is.  Just adding to ROMAN_UNIT_SYMBOLS looks straightforward?
>

Correct, assuming astropy already defines them.
> 
> 
> And just food for thought... some things are unitless but kind of have units.  e.g., you take an image in uncalibrated electrons and multiply it by the flat field with units of calibrated electrons / uncalibrated electrons and get out something with units of calibrated electrons.  I don't think we should do anything there (i.e., I'm not asking for new calibrated / uncalibrated "units"!), but I guess I'm just saying that we should make sure that the reference files, even after adding units, still contain information that explains the content of the files a bit.  Only tangentially related to this PR.

This will involve creating new unit types which is possible within the astropy API, but will be more work.

WilliamJamieson:
> Looks good, modulo code coverage.

The coverage passed? I mean the `UnitMixin` is basically a slight modification of the exact code used by `astropy` it just preserves the types for Roman. If it is a serious issue, I can define a bunch of tests to cover the missed portions of `UnitMixin`

PaulHuwe:
> > Looks good, modulo code coverage.
> 
> The coverage passed? I mean the `UnitMixin` is basically a slight modification of the exact code used by `astropy` it just preserves the types for Roman. If it is a serious issue, I can define a bunch of tests to cover the missed portions of `UnitMixin`

Yeah, the code coverage error'd out instead of providing the comparison, so I can't tell. Mostly noting it because I didn't test them.

ddavis-stsci (approved):
LGTM

PR #110, opened 2022-10-21
Remove `stnode_test`

WilliamJamieson:
Remove the `stnode_test` module from `roman_datamodels` as it appears it does nothing and is not used by anything.


perrygreenfield (approved):
LGTM

PR #111, opened 2022-11-14
Array Quantities

PaulHuwe:
Ticket to support quantities for data arrays in the pipeline.


schlafly (changes requested):
Some of these questions are more general than this PR, but I don't really know what's going on with some of the units, so comments below.
in mk_ramp:
- the border pixel arrays don't have the n_ints dimension, so are they fit as ramps?  In which case the units would be e/s or DN / s or something?  What are they?
- On the other hand the default shape _does_ include an initial (8, ...) for the number of groups, so maybe these should be DN, in which case it looks to me like n_ints isn't being used, and shape is incorrectly documented?

if mk_rampfitoutput:
- What is crmag?  units of counts * s are weird?
- var_poisson and var_rnoise should almost certainly have units of (electron / s)^2
- pedestal should almost certainly have units of electron
- slope, sigslope, yint, sigyint, weights don't have units?

in create_wfi_image:
- var_flat, var_poisson, var_rnoise should probably be (e/s)^2

in create_ramp_fit_output:
- pedestal should probably be in e or DN?
- crmag has different units than in mk_rampfitoutput?
- var_poisson, var_rnoise have e/s instead of (e/s)^2

in create_ramp:
- what is the leading 2 in the dimensions?  This is a 2-resultant exposure?
- what is data?  it's in e/s, so it sounds like a slope, but there are two of them, so I'm confused.

schlafly:
I saw that.  It's a good suggestion.

One nice thing about Quantities is that they allow conversions between equivalent sets of units.  We don't actually have a unit like "m" anywhere, but if we did, I'd be willing to accept either something specified in meters or in centimeters, as long as they have the correct attached unit.  I don't know how annoying it would be to have that flexibility at the schema level.

But it may not be very important to support that since I can't think of very many reasonable sets of units for this data.  I don't think anyone will specify things in terms of kiloelectrons/s.

PaulHuwe:
Converting back to draft.. it appears the units were half done (e.g. mk_level2_image vs mk_rampfitoutput). The perils of code sitting too long between writing and committing.

PaulHuwe:
I have sorted out all the remaining units, and this aligns with RAD-195.

The n_ints questions point to the need for a new ticket: there are n_int variables throughout the file that should be cleaned up (as Roman only has one integration).

src/roman_datamodels/testing/factories.py
```diff
@@ -1067,15 +1084,22 @@ def create_ramp_fit_output(**kwargs):
...
+        "sigyint": _random_array_float32(seg_shape, units=ru.electron),
+        "pedestal": _random_array_float32(seg_shape[1:], units=ru.electron),
+
         "weights": _random_array_float32(seg_shape),
-        "crmag": _random_array_float32(seg_shape),
-        "var_poisson": _random_array_float32(seg_shape),
-        "var_rnoise": _random_array_float32(seg_shape),
+
+        # Magnitude support does not seem to be implemented in ASDF yet
+        # "crmag": u.Magnitude(_random_array_float32(seg_shape) * u.ct / u.s, dtype=np.float32)
```
schlafly, line 1097: We decided this wasn't a mag and have it below; comment can be deleted.

schlafly (approved):
Looks good to me.  I marked one comment as vestigial.

src/roman_datamodels/testing/factories.py
PaulHuwe: Removed.

PR #112, opened 2022-12-05
Remove `enum` enforcement

WilliamJamieson:
With ASDF 2.14.2, the bug in `enum` for schemas as fixed so that we can use `enum` for tagged-string objects in `rad`. This means it is no-longer necessary to enforce these `enum` values through `roman_datamodels`. Thus this PR cleans-up `roman_datamodels` by removing this unnecessary code.

Note that this PR depends on:
- spacetelescope/rad#194 (updates to schemas to re-enable `enum` in those schemas.
- The changes from #111 because spacetelescope/rad#192 was merged and #111 is the support changes for the changes in spacetelescope/rad#192.


nden (approved):
Can point to rad/main now or possible rad release.

WilliamJamieson:
> Can point to rad/main now or possible rad release.

Done

PR #114, opened 2022-12-12
refactor toxenvs, use OpenAstronomy workflows, and move MacOS jobs to schedule

zacharyburnett:
This PR abstracts CI workflows to the OpenAstronomy workflows; this should reduce maintenance for updating and maintaining actions. Additionally, these changes move the majority of MacOS jobs to the weekly scheduled workflow. This should alleviate the issue where the limited MacOS runners for the organization are not available for CI jobs.

image


WilliamJamieson:
@zacharyburnett what is the status of this PR?

zacharyburnett:
> @zacharyburnett what is the status of this PR?

There's an issue with `test-pyargs` that I couldn't pin down, but other than that it looks ready.

WilliamJamieson:
Line 7 of the `ci.yml` workflow should either be removed, or be replaced with a regex like `"*.*.x"` so that it does not match on every single branch (the suggested regex would only match on the release branch pattern).

.github/workflows/ci.yml
```diff
@@ -16,108 +16,46 @@ on:
...
+        - linux: test-numpy120-xdist
+          python-version: 3.9
+        - linux: test-numpy121-xdist
+          python-version: 3.9
+        - linux: test-numpy122-xdist
+          python-version: 3.9
+        - linux: test-numpy121-xdist
+          python-version: 3.10
+        - linux: test-numpy122-xdist
+          python-version: 3.10
```
WilliamJamieson, line 61: Same basic comment as https://github.com/spacetelescope/stcal/pull/153#discussion_r1144948878

WilliamJamieson (changes requested):

pyproject.toml
```diff
@@ -16,7 +16,7 @@ dependencies = [
     'asdf-astropy >=0.3.0',
     'gwcs >=0.18.1',
     'psutil >=5.7.2',
-    'numpy',
+    'numpy >=1.20',
```
WilliamJamieson, line 19: This probably deserves a changelog entry.

WilliamJamieson (approved):

PR #118, opened 2023-01-04
RCAL-439: RDM: Add support for Reference File Quantities

PaulHuwe:
Update factory, maker utils, and tests to support reference file data arrays changed to quantities in RAD.

Also added unit tests for data science quantities.

NOTE: setup.cfg is temporarily pointing to the RAD branch needed for tests to pass. I will update this when the RAD PR is merged.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,8 +29,7 @@ install_requires =
     astropy>=5.0.4
     #rad>=0.14.0
     # Temporary path for github PR tests
-    #rad @git+https://github.com/spacetelescope/rad.git@main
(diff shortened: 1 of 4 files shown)


schlafly (approved):
This all looks good to me.  As I mentioned earlier, though, I'm no longer sure we want to pretend that the linearity coefficients are DN.  But if we think that's the best way to indicate that the input & output units are DN, then I guess that's okay.

ddavis-stsci (approved):
Looks good once the syntax errors are fixed....

PR #121, opened 2023-01-05
Switch to using `ruff` for linting

WilliamJamieson:
Switches the main linter from `flake8` to `ruff`. `ruff` is equivalent to `flake8`, much faster, and supports better configuration options.

Note this PR is based off #121 and should be merged after that PR is merged.

PR #122, opened 2023-01-10
Start using codespell

WilliamJamieson:
Applies the [`codespell`](https://github.com/codespell-project/codespell) to `roman_datamodels` in order to automatically catch (and fix) most common spelling mistakes. Note that if `codespell` catches/fixes a word which is not misspelled, one can easily ignore the word by adding it to a new line under the `ignore-words-list` in the `pyproject.toml` file.

Note this is based on #121 so that it should be merged first.


zacharyburnett (approved):
LGTM when `ruff` PR finishes

PR #123, opened 2023-01-24
RCAL-462: Update RDM for Guidewindow Changes

PaulHuwe:
Ticket to update Guidewindow and Guidestar metadata keywords, and added support for the IPC reference datamodel.

### DIFF
--- a/src/roman_datamodels/testing/factories.py
+++ b/src/roman_datamodels/testing/factories.py
@@ -554,8 +554,10 @@ def create_ipc_ref(**kwargs):
     -------
     roman_datamodels.stnode.IpcRef
     """
+    kernel = np.abs(_random_array_float32((3, 3)))
+    kernel /= np.sum(kernel)
     raw = {
-        "data": _random_array_float32((32, 32)),
+        "data": kernel,
         "meta": create_ref_meta(reftype="IPC"),
     }
     raw.update(kwargs)
--- a/src/roman_datamodels/testing/factories.py
+++ b/src/roman_datamodels/testing/factories.py
@@ -539,6 +539,30 @@ def create_gain_ref(**kwargs):
 
     return stnode.GainRef(raw)
 
+
+def create_ipc_ref(**kwargs):
+    """
+    Create a dummy IpcRef instance with valid values for attributes
+    required by the schema.
+
+    Parameters
+    ----------
+    **kwargs
+        Additional or overridden attributes.
+
+    Returns
+    -------
+    roman_datamodels.stnode.IpcRef
+    """
+    raw = {
+        "data": _random_array_float32((32, 32)),
+        "meta": create_ref_meta(reftype="IPC"),
+    }
+    raw.update(kwargs)
+
+    return stnode.IpcRef(raw)
+
+
 def create_linearity_ref(**kwargs):
     """
     Create a dummy LinearityRef instance with valid values for attributes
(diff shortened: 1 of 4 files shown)


ddavis-stsci (approved):
One minor nit, the GW size for WSM-ACQ-2 is 170x24 pixels.

src/roman_datamodels/testing/factories.py
```diff
@@ -539,6 +539,30 @@ def create_gain_ref(**kwargs):
...
+    ----------
+    **kwargs
+        Additional or overridden attributes.
+
+    Returns
+    -------
+    roman_datamodels.stnode.IpcRef
+    """
+    raw = {
+        "data": _random_array_float32((32, 32)),
```
schlafly, line 558: I don't think it really matters, but we really expect the IPC array to be 3x3, have all positive values, and sum to one.  A slightly more realistic fake IPC kernel would be:
kernel = np.abs(_random_array_float32((3, 3)))
kernel /= np.sum(kernel)

src/roman_datamodels/testing/utils.py
```diff
@@ -701,6 +707,43 @@ def mk_gain(shape=None, filepath=None):
...
+    roman_datamodels.stnode.IpcRef
+    """
+    meta = {}
+    add_ref_common(meta)
+    ipcref = stnode.IpcRef()
+    meta['reftype'] = 'IPC'
+    ipcref['meta'] = meta
+
+    if not shape:
+        shape = (4096, 4096)
```
schlafly, line 735: I think we want (3, 3) here.

src/roman_datamodels/testing/utils.py
```diff
@@ -701,6 +707,43 @@ def mk_gain(shape=None, filepath=None):
...
+
+    ipcref['data'] = np.zeros(shape, dtype=np.float32)
```
schlafly, line 737: For this dummy case, let's do: [[0, 0, 0], [0, 1, 0], [0, 0, 0]].  This is the trivial no-op IPC kernel.

tests/test_models.py
```diff
@@ -312,6 +312,26 @@ def test_opening_gain_ref(tmp_path):
     assert isinstance(gain, datamodels.GainRefModel)
 
 
+# Gain tests
+def test_make_ipc():
+    ipc = utils.mk_ipc(shape=(20, 20))
```
schlafly, line 317: Let's do (3, 3).

schlafly (commented):
A few minor changes before merger to make the IPC kernel somewhat more realistic.

src/roman_datamodels/testing/utils.py
PaulHuwe: Changed.

WilliamJamieson:
@PaulHuwe is this PR the `roman_datamodels` support for spacetelescope/rad#203, and will it fix all the CI failures on `main`, see https://github.com/spacetelescope/roman_datamodels/actions/runs/3997859004?

src/roman_datamodels/testing/utils.py
PaulHuwe: This would break the code were it ever given another shape. Instead I kept the zeros line and added:
`ipcref['data'][int(np.floor(shape[0]/2))][int(np.floor(shape[1]/2))] = 1.0`

tests/test_models.py
PaulHuwe: This should remain a shape different from default to better test the code. The default shape is tested in test_opening_ipc_ref

PaulHuwe:
> @PaulHuwe is this PR the `roman_datamodels` support for [spacetelescope/rad#203](https://github.com/spacetelescope/rad/pull/203), and will it fix all the CI failures on `main`, see https://github.com/spacetelescope/roman_datamodels/actions/runs/3997859004?
> 
> ASDF is picking up these failures in its downstream testing: [asdf-format/asdf#1329 (comment)](https://github.com/asdf-format/asdf/pull/1329#issuecomment-1402115795)

Yes, amongst other RAD updates. This ticket is supposed to get RDM completely up to date with RAD.

tests/test_models.py
schlafly: Can we at least make it odd?  With an even convolution kernel shape I feel like we also need to define how it's centered.

PR #124, opened 2023-01-27
adding source_detection

cshanahan1:
The source detection step adds a meta.source_detection section, in which there is an attribute for the filename, and the associated source catalog (an astropy Table).


cshanahan1:
closing this, replaced with #132

PR #128, opened 2023-03-01
RCAL-436 Move Maker Utilities

PaulHuwe:
Moving datamodel maker utilities from testing/utils to the main directory in order to better reflect their usage. As part of that,  also split random functions out to their own utility file and linking to them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 
 - Added support for Inverse Nonlinearity data model, maker utilities, and tests. [#125]
 
+- Moved datamodel maker utilities and split random functions out to utility file. [#128]
+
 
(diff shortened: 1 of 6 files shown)


schlafly:
This looks good to me.  What do we expect it to break?
- romancal?
- INS wfisim wrapper
- simulator
- ?

PaulHuwe:
That's the list for packages. Everything else would be people's personal preferences, any old instructional notebooks lying around, etc.

mairanteodoro (approved):
Looking good to me.

ddavis-stsci (approved):
Looks good.

schlafly (approved):

PR #131, opened 2023-03-09
Deprecate `roman_datamodels.units`

WilliamJamieson:
astropy/asdf-astropy#142 introduced the ability to serialize non-VOunits defined within `astropy` to a special ASDF tag. All of the non-VOunits needed by Roman fall under this category. Because serialization of non-VOunits was previously possible, #109 created a wrapper "hack" around the `astropy` non-VOunits needed by Roman to enable support for serialization to a special ASDF tag created in `rad` (see spacetelescope/rad#168).

This PR in conjunction with spacetelescope/rad#220, switches `roman_datamodels` to directly use the non-VOunit support now in `asdf-astropy` for writing new Roman files.

Note @schlafly requested, that we continue support for the ability to read Roman files written with the Roman non-VOunit tags because right now it would be burdensome to have to regenerate all the current Roman files. This PR specifically makes efforts to preserve the ability to read/validate these existing files. However, this support is limited only to reading these tags. Once this PR is merged, no new files will be written including the Roman non-VOunit tag, and any files read/resaved will have these tags changed to the astropy ones.

This means that this PR is an intermediate solution until we reach a point where it is convenient to regenerate all existing Roman files, at which time the remaining parts of the `roman_datamodels.units` can be removed.

Additionally, a deprecation warning has been added to the `roman_datamodels.units` module which now only re-exports the `electron` and `DN` units from `astropy.units` directly. This is to allow some flexibility to those who have gotten used to using the `roman_datamodels` forms of those units.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@
 
 - Moved datamodel maker utilities and split random functions out to utility file. [#128]
 
+- Begin process of removing ``roman_datamodels.units`` for non-VOUnit support in favor
+  of non-VOUnit support coming directly via ``asdf-astropy``. [#131]
+
 
 0.14.1 (2023-01-31)
 ===================
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,12 +13,12 @@ classifiers = [
 ]
 dependencies = [
     'asdf >=2.14.2',
-    'asdf-astropy >=0.3.0',
+    'asdf-astropy >=0.4.0',
     'gwcs >=0.18.1',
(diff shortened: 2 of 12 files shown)


schlafly (approved):
Looks good to me.

ddavis-stsci (approved):
LGTM

PaulHuwe (approved):
LGTM - I love seeing all the code streamlining. I have confirmed that this passes all tests throughout the pipeline (including RCAL regression tests).

PR #135, opened 2023-03-19
Update test that will begin failing in a future version of asdf

eslavich:
There's a bug of some kind in asdf that I accidentally fixed today, which will cause this test to fail when released.  If I understand correctly, the validate-on-assignment feature is supposed to be causing these statements to fail:

```python
af.tree["roman"].meta.telescope = "NOTROMAN"
```
```python
af.tree["roman"].meta.origin = "NOTSTSCI"
```

... but they do not under currently released versions of asdf.  This change will allow the test to pass both before and after the bug is fixed.

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -65,17 +65,24 @@ def test_core_schema(tmp_path):
         af.tree = {"roman": wfi_image}
 
         # Test telescope name
-        af.tree["roman"].meta.telescope = "NOTROMAN"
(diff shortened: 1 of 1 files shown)


WilliamJamieson (approved):
Hey you fixed that bug. It was on a list of minor bugs that needed to be fixed when I had time.

I was always curious about why that was happening... well I'm glad that was fixed

PR #140, opened 2023-03-29
when calling Datamodel.validate use a schema if available

braingram:
fixes RCAL-538

A call to Datamodel.validate results in a call to _check_value which appears to use a default schema to check that a new value is valid.
https://github.com/spacetelescope/roman_datamodels/blob/36b556ec83c2f5530a78237bd6669739383bf4c9/src/roman_datamodels/validate.py#L64-L69

However, as described in RCAL-538, an issue can arise if sub-contents of a node are changed (for example if meta is accessed like a dictionary and a value is changed, see the test added to this PR). This sub-content change will not trigger validation (as there's no way to know that the sub-contents have changed). Furthermore, the default schema provided during the call to validator.validate above will take precedence over any tag associated schema. This results in failure to catch invalid objects if the sub-content was changed (probably only for the first children but I did not test this).

This PR addresses the issue by looking for an available schema when validate is called which will be used in place of the temp_schema (the temp_schema will still be used if no instance schema is found).

### DIFF
--- a/src/roman_datamodels/validate.py
+++ b/src/roman_datamodels/validate.py
@@ -33,7 +33,7 @@ def value_change(value, pass_invalid_values, strict_validation):
         if pass_invalid_values:
             update = True
         if strict_validation:
-            raise jsonschema.ValidationError(errmsg)
+            raise errmsg
         else:
             warnings.warn(errmsg, ValidationWarning)
     return update
@@ -61,7 +61,10 @@ def _check_value(value):
 
     validator_context = AsdfFile()
 
-    temp_schema = {"$schema": "http://stsci.edu/schemas/asdf-schema/0.1.0/asdf-schema"}
+    if hasattr(value, "_schema"):
+        temp_schema = value._schema()
+    else:
+        temp_schema = {"$schema": "http://stsci.edu/schemas/asdf-schema/0.1.0/asdf-schema"}
     validator = asdf_schema.get_validator(temp_schema, validator_context, validators=validator_callbacks)
 
     value = yamlutil.custom_tree_to_tagged_tree(value, validator_context)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@
 
 - update minimum version of ``numpy`` to ``1.20`` and add minimum dependency testing to CI [#114]
 
+- Use available tag schema if available during datamodels.validate [#140]
+
 0.14.1 (2023-01-31)
 ===================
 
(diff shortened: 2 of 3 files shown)


ddavis-stsci (approved):
My tests show that this works as I'd expect and yields the same results as the .save method. 

2023-03-29 14:07:11,568 - stpipe.SourceDetectionStep - INFO - Step SourceDetectionStep done
>>> result.validate()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.10/site-packages/roman_datamodels/datamodels.py", line 303, in validate
    validate.value_change(self._instance, pass_invalid_values=False, strict_validation=True)
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.10/site-packages/roman_datamodels/validate.py", line 36, in value_change
    raise errmsg
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.10/site-packages/roman_datamodels/validate.py", line 29, in value_change
    _check_value(value)
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.10/site-packages/roman_datamodels/validate.py", line 71, in _check_value
    validator.validate(value, _schema=temp_schema)
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.10/site-packages/jsonschema/validators.py", line 314, in validate
    raise error
jsonschema.exceptions.ValidationError: mismatched tags, wanted 'asdf://stsci.edu/datamodels/roman/tags/source_detection-1.0.0', got 'tag:yaml.org,2002:map'

Failed validating 'tag' in schema['properties']['meta']['allOf'][1]['properties']['source_detection']:
    {'tag': 'asdf://stsci.edu/datamodels/roman/tags/source_detection-1.0.0'}

On instance['meta']['source_detection']:
    {'tweakreg_catalog': {'byteorder': 'little',
                          'datatype': 'float64',
                          'shape': [4, 3833],
                          'source': 8}}

src/roman_datamodels/validate.py
```diff
@@ -54,14 +54,17 @@ def _check_type(validator, types, instance, schema):
 validator_callbacks.update({"type": _check_type})
 
 
-def _check_value(value):
+def _check_value(value, schema_uri=None):
```
WilliamJamieson, line 57: The `schema_uri` argument is not used and can be removed.

WilliamJamieson (changes requested):

src/roman_datamodels/validate.py
braingram: Thanks for the catch! I removed the unused argument.

WilliamJamieson (approved):

PR #148, opened 2023-04-13
Refactor `random_utils` to allow seeding

### LINKED ISSUE #144: random_utils need to let user set seed
I don't see any way to set a seed for https://github.com/spacetelescope/roman_datamodels/blob/main/src/roman_datamodels/random_utils.py 

Also while you are at it, maybe use the new API: https://numpy.org/doc/stable/reference/random/generator.html

WilliamJamieson:
Refactor's `random_utils` to only use `numpy.random` for random number generation, and enable users to seed random numbers.

Fixes #144

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@
 
 - Remove ``source_type_apt`` from ``target-1.0.0`` related datamodels. [#152]
 
+- Enable seeding for ``random_utils`` functions. [#148]
+
 0.14.2 (2023-03-31)
 ===================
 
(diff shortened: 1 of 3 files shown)


WilliamJamieson:
Regression tests pass, see: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/135/

PaulHuwe (approved):
LGTM

pllim:
Thanks!

I find it interesting that non of your existing test files here need their answers updated. But I guess that also makes sense since before this, the fake data was completely random...

Issue #151, opened 2023-04-13
Inconsistent fields between factories and `rad` schemas


WilliamJamieson:
This test: https://github.com/spacetelescope/roman_datamodels/blob/626e0da697614cefaef3b26d06fc3f528e2c7dab/tests/test_factories.py#L30-L46 is currently being skipped by our unit tests.

I believe that this is a bug in the parameterization, where `isinstance` should be `issubclass`, i.e. correcting line 30 to:
```python
@pytest.mark.parametrize("node_class", [c for c in stnode.NODE_CLASSES if issubclass(c, stnode.TaggedObjectNode)])
```

This then runs `test_no_extra_fields` on all the `datamodel` objects constructed from the `rad` schemas. Almost all of the `factories` pass this test. However, there are 3 models that do not pass:

- `Pixelarea`: extra keys: `meta`
- `Target`: extra keys: `source_type_apt`
- `RefFile`: extra keys: `distortion`

Are these errors in the `factories` for these classes, or are they missing fields in the `rad` schemas which should be added?

WilliamJamieson:
The `Pixelarea` object corresponds to this schema:
> https://github.com/spacetelescope/rad/blob/c87ae34f3e99362b8a370431c10dabbbee732619/src/rad/resources/schemas/pixelarea-1.0.0.yaml#L1-L18

Which seems to be a vestigial schema from before the `reference_files` where moved into a separate set of schemas. I believe that the schema in question was superseded by:
> https://github.com/spacetelescope/rad/blob/c87ae34f3e99362b8a370431c10dabbbee732619/src/rad/resources/schemas/reference_files/pixelarea-1.0.0.yaml#L1-L42

So I believe the `pixelarea-1.0.0` schema should be removed (not to be confused with the `reference_files/pixelarea-1.0.0` schema).

WilliamJamieson:
The missing `distortion` key in `RefFile` I believe is an oversight in `rad`, as all the `reference_files` schemas are given a designation in the [`ref_file-1.0.0`](https://github.com/spacetelescope/rad/blob/c87ae34f3e99362b8a370431c10dabbbee732619/src/rad/resources/schemas/ref_file-1.0.0.yaml#L37-L63) schema except `distortion.

WilliamJamieson:
[...]

PR #152, opened 2023-04-13
Remove the `source_type_apt` keyword

WilliamJamieson:
spacetelescope/rad#206 removed `source_type_apt` from the `target-1.0.0` schema. This change was never made in `roman_datamodels` and was discovered in issue #151.

This PR is the matching PR to spacetelescope/rad#206 to remove `source_type_apt` from `roman_datamodels`


nden (approved):
One typo needs to be fixed, otherwise looks good.

PR #154, opened 2023-04-14
RCAL-543: Read Pattern Support

PaulHuwe:
Add support for read pattern in data model makers and factories.

PLWishMaster link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/127/


schlafly:
For these read_pattern defaults I think we should pick something that is at least a little consistent with the other exposure stanza metadata.  I forget what our default "ngroups" is, but it feels bad if that's not consistent with the read_pattern; let's at least match those two?

PaulHuwe:
> For these read_pattern defaults I think we should pick something that is at least a little consistent with the other exposure stanza metadata. I forget what our default "ngroups" is, but it feels bad if that's not consistent with the read_pattern; let's at least match those two?

Applied and added to the test.

PR #155, opened 2023-04-14
drop support for Python 3.8

zacharyburnett:
Numpy is imminently dropping support for Python 3.8

PR #156, opened 2023-04-18
docs build - fix for `automodapi`

zacharyburnett:
apply fix from https://github.com/spacetelescope/roman_datamodels/pull/155#issuecomment-1513371214 for docs build issue in `Python > 3.8`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,7 @@ else:
 
 def find_mod_objs_patched(*args, **kwargs):
     from sphinx_automodapi.utils import find_mod_objs
-    
+
     return find_mod_objs(args[0], onlylocals=True)
 
 


zacharyburnett:
It looks like the build fails with 
https://readthedocs.org/projects/roman-datamodels/builds/20154238/
```
TypeError: <lambda>() got an unexpected keyword argument 'onlylocals'
```

nden:
It looks like the behavior has changed with Python 3.9. The issue and a proposed updated solution is documented [here](https://github.com/astropy/sphinx-automodapi/issues/119).

ddavis-stsci (approved):
Looks OK to me, hopefully they will update the astropy/sphinx code so we can remove the patch at some point.

nden:
The failure is unrelated.

WilliamJamieson:
Short answer is fix `stuserdict.py`, which is what is causing the issue. This can be done in several ways (note these are not mutually exclusive improvements):

1. Define an `__all__ = ["STUserDict"]` in `stuserdict.py`. Per [pep-8](https://peps.python.org/pep-0008/#public-and-internal-interfaces):

> To better support introspection, modules should explicitly declare the names in their public API using the `__all__` attribute. Setting `__all__` to an empty list indicates that the module has no public API.

So `__all__` should be defined anyways

2. The issue is it attempting to generate documents for `Any` and `Iterator` which are used purely as type annotations. This is because they are imported as no private things in the namespace there. As they are purely type annotations changing: https://github.com/spacetelescope/roman_datamodels/blob/52b54519c1572d4907f7014afc5ce9bf52f73ff6/src/roman_datamodels/stuserdict.py#L1 to
```python
from __future__ import annotations
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from typing import Any, Iterator
from typing import Any, Iterator
```
Allows for the type-hints to say, but they will only be imported if type checking via the annotations is used.

3. Remove the type hints all together. No where else in `roman_datamodels` are there type hints.

The either options 1 and 2 or 1 and 3 would be the best way to deal with the issue.

Note, `STUserDict` seems redundant and should be removed. It is only a base class for `stnode.DNode`, which overloads much of its interface anyways. Moreover, `STUserDict` appears to be a copy-paste from the CPython standard library's `UserDict` (from 3.8/3.9) where `.data` has been replaced by `._data`. Instead it would be simpler to simply have `DNode` subclass `collections.abc.MutableMapping` directly. However, this type of change is outside the scope of this bugfix.

PR #165, opened 2023-04-26
RCAL-435: Renamed n_ints to n_groups. Did some  cleanup.

PaulHuwe:
Resolves [RCAL-435](https://jira.stsci.edu/browse/RCAL-435)

Closes #

This PR corrects n_ints to n_groups in maker utilities. In addition, it cleans up some default shape assignment.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/142/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,9 @@
 
 - Add Pull Request Template. [#147]
 
+- Renamed n_ints to n_groups. Did some shape variable cleanup. [#165]
+
+
 0.14.2 (2023-03-31)
 ===================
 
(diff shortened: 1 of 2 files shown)


schlafly:
I agree that it's good to get rid of n_ints here.  We expect to some day use n_resultants rather than n_groups, though, so you may need to do this again in the future, but thumbs up until we pull the trigger there.

ddavis-stsci (approved):
LGTM
Won't this eventually be n_resultants? or whatever we agree to with INS?

PaulHuwe:
Eventually, yes, but for now this gets rid of the anachronistic n_ints.

PR #170, opened 2023-05-08
Fix `patternProperties` handling

WilliamJamieson:
This test should pass when the issue is fixed.
Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes #

This pull request fixes two related bugs:

1. [`Dnode._schema()`
](https://github.com/spacetelescope/roman_datamodels/blob/ae1caffaa0cccebdbfa2ac41907c8e6717d52ad2/src/roman_datamodels/stnode.py#L224-L233) is assumed to return the schema for the `Dnode` instance within itself when it attempts to get the `parent` schema. This prevents the setting of properties below the "main" schema level via `Dnode.foo = bar`. 
2. [`_get_schema_for_property`](https://github.com/spacetelescope/roman_datamodels/blob/ae1caffaa0cccebdbfa2ac41907c8e6717d52ad2/src/roman_datamodels/stnode.py#L102-L111) cannot handle any schema which includes `patternProperties`.

The presence of these bugs means that for the `WfiImgPhotomRefModel` setting things like `photmjsr` becomes difficult, and fails to validate the [schema](https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml#L25-L29) when it is set. e.g.

```python
from roman_datamodels.testing.factories import create_wfi_img_photom_ref

mdl = create_wfi_img_photom_ref()

# Fails due to bug 1
mdl.phot_table.F062.photmjsr = 3.14

# When bug 1 is fixed, this does not raise a ValidationError due to bug 2
mdl.phot_table.F062.photmjsr = 3.14
```

Moreover, fixing this issue also points out that the `__setitem__` api really should be running validation too. e.g.
```
# Should raise ValidationError
mdl["phot_table"]["F062"]["photmjsr"] = 3.14
```

This PR fixes the two bugs and adds the additional validation step, which will hopefully prevent outside users from creating models which cannot be written.

**Checklist**
- [ ] updated relevant tests


WilliamJamieson:
> setitem should not throw and error if the value is not allowed by the schema. It should be a warning or be optional. 

Why? This makes no sense and is asking for trouble. Any use of __setitem__ to store data in violation of the schema is an abuse of the DNode/datamodel object and should not occur. This is because there is an implicit assumption that these objects (or the objects contained within) are defined via the schema they were constructed from; meaning that, setting invalid data in the object is asking for bugs to occur. In particular this is especially true for DNode objects, since DNode objects are dynamically created at runtime. This means developers can only reliably understand the object's content via examination of the schema.

I also disagree with defaulting to a warning (instead of an error), because development thus far seems to care little about whether or not warnings are issued. Effectively, this puts us back at my original issue; namely, when someone in the future is dealing with invalid data problem when they go to fix such warnings. Instead such bugs should be addressed from the beginning. 

If we really must have an interface to enable setting without validation, a context manager should be employed to temporarily suspend data validation. When the manager exits it can then call validation on the object in question. Alternatively, an explicit method such as set_without_validation could be added to enable setting data without validation.

Either of the above interfaces become necessary because the __setitem__ interface does not allow for "optional" arguments in the traditional sense. So there is not a way to inject such an "optional" argument using the supporting Python syntax.

In any case, I do agree that such an interface addition/change is outside the scope of this PR. So I will open a separate issue/PR concerning this.

perrygreenfield:
@ddavis-stsci Can you provide a compelling case for why values shouldn't be validated?

ddavis-stsci:
So we're effectively saying that if IPAC uses this for code development 
then they need to setup/update a rad schema
for every change even if they don't end up using it. Currently when we 
do development we can set validate_on_write to
False and develop the code with the test asdf files. This seems to break 
that model.

In my opinion this would make having other entities using this for code 
development awkward at best.

perrygreenfield:
So you are basically saying that in the development process it is a hassle as one is making changes (not knowing necessarily ahead of time the final target) to keep tracking a schema to each change or addition? If so, would some global setting that turns this checking off work for development?

ddavis-stsci:
I think that basically sums it up. In production having strict checks 
can be beneficial in development not so much.

I'd prefer a setting that would turn that on in the operational 
environment and the default to be permissive, at least
for now.

ddavis-stsci:
If we were later in the process I would agree. At this point we don't 
have any users (or few, I&T for example) that are not
basically doing development. While we are trying to get folks to adopt 
and use our code I'd advocate to make it as
easy as possible. I'm worried about someone trying to set an attribute 
and getting an error, and getting frustrated.

Yes, they should read the documentation first. If I could enforce that ...

WilliamJamieson:
> So we're effectively saying that if IPAC uses this for code development then they need to setup/update a rad schema for every change even if they don't end up using it. Currently when we do development we can set validate_on_write to False and develop the code with the test asdf files. This seems to break that model. In my opinion this would make having other entities using this for code development awkward at best.

I think this isn't as big a problem as you believe. Presumably, IPAC's contributions will mostly be on their own schemas, which means they can simply set their backing schema to be completely permissive. That is just specify it as an object without any underpinning structure. That will allow them to treat the model entirely as a dictionary until they settle on how they want to formally specify the model. Or even more broadly, they can simply fail to specify a schema for the `tag` they are using, in which case validation is entirely skipped.

I am completely okay with letting (even encouraging) IPAC (or even STScI) to contribute completely permissive schemas to RAD for development purposes. These can be tracked and refined to final strict forms as development proceeds.

> If we were later in the process I would agree. At this point we don't have any users (or few, I&T for example) that are not basically doing development. While we are trying to get folks to adopt and use our code I'd advocate to make it as easy as possible. I'm worried about someone trying to set an attribute and getting an error, and getting frustrated. Yes, they should read the documentation first. If I could enforce that ...

I completely disagree with this. While early in the development we should in fact constrain the code as much as possible. Allowing it to be permissive and then tightening the process will just lead to us breaking these users code later on. It is a far more sustainable process (for developers and downstream users) to strictly limit the API and then minimally loosen it when circumstances require us to do so. This will lead to less headaches in the future.

I believe we should make the API very explicit to the user about what they are doing rather than having an implicit work-around. IMO it is more frustrating for users to "discover" some side channel (`__setitem__` in this case) which circumvents some "error" (validation error) without that side channel explicitly (via an explicitly named context manager or bound method) only to discover that this causes them other problems later on (such as having to use `validate_on_write=False` when writing files). It is more honest to users to raise an explicit (and informative message), than it is to let them shoot themselves in the foot.

> I probably would have it the other way, strict checks by default. I don't see it as a big problem for developers to disable that in their development code. (Actually, the reality is that operational environment may have validation off as well since it is presumed that tests have passed, and it may only be turned on to debug a failure case; it is for the testing and outside users that the default case is most important.)

I agree that this is a much better route. Supplying a context manager to "circumvent" validation, leads us having a "roman datamodels config" (essentially a "global context"). `roman_datamodels` already has several "globalish" control variables as it is, see https://github.com/spacetelescope/roman_datamodels/blob/ae1caffaa0cccebdbfa2ac41907c8e6717d52ad2/src/roman_datamodels/stnode.py#L47-L48 and https://github.com/spacetelescope/roman_datamodels/blob/ae1caffaa0cccebdbfa2ac41907c8e6717d52ad2/src/roman_datamodels/stnode.py#L54-L56z
for a few examples. Ideally, these would instead be stashed within the "roman datamodels config" and be directly modifiable and again allow context managed modification.

Error messages can clearly indicate (link to) documentation which explicitly lays out how to handle this.

ddavis-stsci (changes requested):
I don't see the regression tests for this change.

WilliamJamieson:
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/172/

ddavis-stsci (changes requested):
It looks like the romancal regression tests are passing but I'm seeing errors in the roman_datamodels unit tests,

```
pytest tests/test_stnode.py
=================================================================== test session starts ===================================================================
platform darwin -- Python 3.11.0, pytest-7.3.1, pluggy-1.0.0
rootdir: /Users/ddavis/src/Roman/roman_datamodels
configfile: pyproject.toml
plugins: remotedata-0.4.0, hypothesis-6.75.1, asdf-2.15.0, mock-3.10.0, filter-subpackage-0.1.2, astropy-header-0.2.2, doctestplus-0.12.1, astropy-0.10.0, cov-4.0.0, ci-watson-0.6.1, openfiles-0.5.0, arraydiff-0.5.0
collected 51 items

tests/test_stnode.py ..................................................F                                                                            [100%]

======================================================================== FAILURES =========================================================================
_______________________________________________________________ test_set_pattern_properties _______________________________________________________________

    def test_set_pattern_properties():
        """
        Regression test for `photmjsr` not being validated because it is a patternProperty
        """

        # This model uses the `photmjsr` patternProperty
        mdl = factories.create_wfi_img_photom_ref()

        with pytest.raises(asdf.ValidationError):
            # This should be invalid because it is not a quantity
>           mdl.phot_table.F062.photmjsr = 3.14

tests/test_stnode.py:95:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../miniconda3/envs/rcal_python311/lib/python3.11/site-packages/roman_datamodels/stnode.py:174: in __setattr__
    self._schema()
../../../miniconda3/envs/rcal_python311/lib/python3.11/site-packages/roman_datamodels/stnode.py:232: in _schema
    subschema = _get_schema_for_property(parent_schema, self._name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

schema = None, attr = 'F062'

    def _get_schema_for_property(schema, attr):
>       subschema = schema.get("properties", {}).get(attr, None)
E       AttributeError: 'NoneType' object has no attribute 'get'

../../../miniconda3/envs/rcal_python311/lib/python3.11/site-packages/roman_datamodels/stnode.py:103: AttributeError
================================================================= short test summary info =================================================================
FAILED tests/test_stnode.py::test_set_pattern_properties - AttributeError: 'NoneType' object has no attribute 'get'
============================================================== 1 failed, 50 passed in 23.77s
```

WilliamJamieson:
> It looks like the romancal regression tests are passing but I'm seeing errors in the roman_datamodels unit tests,
> 
> ```
> pytest tests/test_stnode.py
> =================================================================== test session starts ===================================================================
> platform darwin -- Python 3.11.0, pytest-7.3.1, pluggy-1.0.0
> rootdir: /Users/ddavis/src/Roman/roman_datamodels
> configfile: pyproject.toml
> plugins: remotedata-0.4.0, hypothesis-6.75.1, asdf-2.15.0, mock-3.10.0, filter-subpackage-0.1.2, astropy-header-0.2.2, doctestplus-0.12.1, astropy-0.10.0, cov-4.0.0, ci-watson-0.6.1, openfiles-0.5.0, arraydiff-0.5.0
> collected 51 items
> 
> tests/test_stnode.py ..................................................F                                                                            [100%]
> 
> ======================================================================== FAILURES =========================================================================
> _______________________________________________________________ test_set_pattern_properties _______________________________________________________________
> 
>     def test_set_pattern_properties():
>         """
>         Regression test for `photmjsr` not being validated because it is a patternProperty
>         """
> 
>         # This model uses the `photmjsr` patternProperty
>         mdl = factories.create_wfi_img_photom_ref()
> 
>         with pytest.raises(asdf.ValidationError):
>             # This should be invalid because it is not a quantity
> >           mdl.phot_table.F062.photmjsr = 3.14
> 
> tests/test_stnode.py:95:
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> ../../../miniconda3/envs/rcal_python311/lib/python3.11/site-packages/roman_datamodels/stnode.py:174: in __setattr__
>     self._schema()
> ../../../miniconda3/envs/rcal_python311/lib/python3.11/site-packages/roman_datamodels/stnode.py:232: in _schema
>     subschema = _get_schema_for_property(parent_schema, self._name)
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> 
> schema = None, attr = 'F062'
> 
>     def _get_schema_for_property(schema, attr):
> >       subschema = schema.get("properties", {}).get(attr, None)
> E       AttributeError: 'NoneType' object has no attribute 'get'
> 
> ../../../miniconda3/envs/rcal_python311/lib/python3.11/site-packages/roman_datamodels/stnode.py:103: AttributeError
> ================================================================= short test summary info =================================================================
> FAILED tests/test_stnode.py::test_set_pattern_properties - AttributeError: 'NoneType' object has no attribute 'get'
> ============================================================== 1 failed, 50 passed in 23.77s
> ```

It appears that you do not have a current version of this code. The CI run automatically on this PR does not detect this failure; moreover, this is literally the result of issue (1) listed in the description and it has been fixed by:

https://github.com/WilliamJamieson/roman_datamodels/blob/f93f3e570305bceb7c47b8212a535c1c217e98ef/src/roman_datamodels/stnode.py#L227

I also cannot reproduce this error locally, so you will need to provide more information on how you are seeing this failure.

WilliamJamieson:
I'm not doubting that you have some version of my branch because you are seeing the failure of the test I wrote which reproduces the failures I was seeing before my fixes.

Since the CI runs without issue, I strongly suspect it is a local environment issue on your end. There are two likely possibilities:

1. You have an older version of my branch. I purposely structure my bugfix changes so that the first commits introduce unit tests which reproduce the failure followed by commits which work to fix those tests.
2. Your branch is up to date, but your local Python has not installed my branch's version of `roman_datamodels`. This would pick up this failure because `pytest` will run the tests in the local files you are looking at, but those files import from `roman_datamodels`. If your local Python environment is not configured to use the version of `roman_datamodels` in my branch, this effectively recreates the reproducer environment prior to my fixes. I suggest you either rebuild your local environment or at the very least run `pip install -e .` in your repository clone.

tests/test_stnode.py
```diff
@@ -79,3 +80,18 @@ def test_schema_info():
...
+    """
+    Regression test for `photmjsr` not being validated because it is a patternProperty
+    """
+
+    # This model uses the `photmjsr` patternProperty
+    mdl = factories.create_wfi_img_photom_ref()
+
+    with pytest.raises(asdf.ValidationError):
+        # This should be invalid because it is not a quantity
+        mdl.phot_table.F062.photmjsr = 3.14
```
WilliamJamieson, line 95: @nden this part of the test should indicate that your issue has been resolved.

As for setting it via the `__setitem__` interface, @ddavis-stsci does not want that interface to run validation. So I cannot prevent that interface from allowing any type of data to be set.

WilliamJamieson:
Note that the CI will experience some failures related to the removal of the unit schema until #172 is merged.

ddavis-stsci (approved):

PR #176, opened 2023-05-11
Rename `datamodels.open` to avoid builtin shadowing

WilliamJamieson:
Currently the `datamodels` module defines a function `open`, which of course shadow's Python's open function. In general, it is best practice to avoid name shadowing like this. Model container needs to be able to use the builtin open so it has to perform a kludge to do this.

This PR resolves all of this by renaming `open` to `roman_open` in the `datamodels` module and then exports `roman_open` under 	`open` in the `__init__.py` for all of `roman_datamodels`. This effectively means anyone importing from `roman_datamodels.open` should not have any issues.

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,10 @@
 
 - Refactor the ``maker_utils`` to be easier to maintain and test. [#193]
 
+- Rename ``open`` in the ``datamodels`` module to ``roman_open`` to avoid
+  shadowing the builtin ``open`` function, while preserving ``roman_datamodels.open``
+  in the main namespace. [#176]
+  
 0.15.0 (2023-05-15)
 ===================
 
@@ -39,6 +43,7 @@
 - Add support for model containers constructed from ``Iterable`` [#164]
 
 - drop support for Python 3.8 [#155]
+
 
 
 0.14.2 (2023-03-31)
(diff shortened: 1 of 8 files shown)


mairanteodoro (approved):
Looks good to me. Thanks, @WilliamJamieson!

PaulHuwe:
Can you add regression test link?

nden:
I'd like to take a step back and ask how do people usually use this. I usually run

```
from roman_datamodels import datamodels as rdm
rdm.open(...)
```
as well as

```
from jwst import datamodels
datamodels.open(...)
```

What do others use?
What was the original concern - using `roman_datamodels.datamodels.open` or using `roman_datamodels.open`?

I think we need to decide on a convention and recommend it.  It'd be good if the interface is consistent to some extent with the one for jwst. `from roman_datamodels import datamodels as rdm` was what I considered the current convention.

mairanteodoro:
> I'd like to take a step back and ask how do people usually use this. I usually run
> 
> ```
> from roman_datamodels import datamodels as rdm
> rdm.open(...)
> ```
> 
> as well as
> 
> ```
> from jwst import datamodels
> datamodels.open(...)
> ```
> 
> What do others use? What was the original concern - using `roman_datamodels.datamodels.open` or using `roman_datamodels.open`?
> 
> I think we need to decide on a convention and recommend it. It'd be good if the interface is consistent to some extent with the one for jwst. `from roman_datamodels import datamodels as rdm` was what I considered the current convention.

The original concern was that we couldn't use the built-in `open` function inside `roman_datamodels.datamodels.py` because of the name shadowing.

mairanteodoro:
> Can you add regression test link?

To run the regression test, we'd have to add a `JenkinsfileRT` to this project since it doesn't have one yet.

nden:
Let's see if this is needed after the new ModelContainer is merged into romancal.

PR #181, opened 2023-05-16
Add type stubs for `stnode` objects.

WilliamJamieson:
# Description
Type stubs are (`*.pyi`) files described by [PEP561](https://peps.python.org/pep-0561/) which enable the indication of type hints for Python in a file external to the file they are defined in. This is useful when code "exists" which is not type hinted but which the author either does not have control over to add hints, or does not want to clutter the main code base.

There are a few advantages to adding type stubs:

1. They give an IDE (or code completion program/plugin) additional information to assist with giving helpful completion options to its user (the programmer).
2. They enable better static code analysis by analysis tools by allowing for (limited) static type checking.

NOTE: type hints and/or type stubs are not enforced at runtime by Python directly (there do exist tools which do perform this, but I have no intention of implementing such tools).

This pull request adds type stubs which describe all of the types created by `stnode` dynamically at runtime based on the RAD schemas. Additionally, it includes checks that these stubs are consistent with the RAD schemas and methods to automatically generate sample instances of each of these objects.

Why do I want to do this?

1. Currently, the types created dynamically by `stnode` at first glance "appear out of thin air". So first inspection of `roman_datamodels` indicates that the types in question do not exist. This is confusing to some users/contributors at times. The type stubs act to document these "dynamically" created objects so that users/contributors have a python based reference to look at when working with these stubs. IMOP it is much easier to read the type stubs than it is to go to RAD and parse the schemas in order to figure out what things correspond to what in the dynamic objects.
2. The confusion extends to the IDE tooling I mentioned above, in that it cannot generate completion hints for the user while they are writing code. Particularly, when they are first working with one of these objects in a given context (usually some `*.py` file) the IDE will be of little use because it will fall back on completions based on code within the file being edited. Most modern IDEs understand what type stubs are and how to use them. This means with the type stubs give the IDE enough information to make very good suggestions.
3. Similarly, the type stubs enable the IDE to flag some invalid variable assignments. Since the `stnode` objects all perform  some level of validation at runtime of the data set inside them, it can be frustrating to get runtime errors because the data being set is "invalid". Static type checking by the IDE can alleviate some of these issues.
4. It gives a much "cleaner" method of generating data like the "factory" methods (or even the `maker_utils`) because it does not require as much ad hoc modification of code when things change. Instead it moves it to simply updating the type stubs to reflect RAD changes. (It might even be possible generate the type stubs automatically from RAD, but I have not attempted this yet).

In summary, this should make `roman_datamodels.stnode` way nicer to work with in the future.

Why does this work?

- The "dynamically" created objects by `stnode` are really not that "dynamic" in the sense that `stnode` will always generate the same objects every time `roman_datamodels` is run for any fixed version of RAD. The dynamic creation is just a function of translating the schema description into a python description that we can use, serialize to asdf, and deserialize from asdf into. So in a real sense they are "static", but we abuse python to make things easier on ourselves.
- Technically, all the `stnode` objects are at their heart just dictionaries, but `__getattr__` and `__setattr__` methods exist to turn the "keys" transparently into object attributes backed by a validation system. The type stubs  do not actually document this. Instead they pretend that all these key/value pairs are just object attributes, so that the completion systems have an easier time dealing with them. This also makes them much clearer to users attempting  to understand the type's construction. So the type stubs technically "lie" about the object, but this is transparent to developers and the IDE because of the `__getattr__`/`__setattr__` interface.
- I had to hack Python a bit to get testing to properly work in that Python really doesn't want to load stub files at runtime, which makes it difficult to test their integrity. However, all stub files have to be valid python  in their own right, so I used a symbolic link to change the `.pyi` to `.py` so that Python would just recognize the file and load it properly so testing and use of instance creation routines could be exploited.

NOTE: I can also move the type stubs into an entirely separate package from `roman_datamodels` and just declare it stubs the `roman_datamodels` package. Python will link the type stubs to the associated code when both packages are installed. So this code does not have to go directly in `roman_datamodels`. However, I would prefer if it did for ease of maintenance.  

**Checklist**
- [ ] updated relevant tests


src/roman_datamodels/datamodels.py
```diff
@@ -16,10 +16,10 @@
...
 from pathlib import PurePath
 
-import asdf
-import numpy as np
-import packaging.version
-from astropy.time import Time
+import asdf  # type: ignore
+import numpy as np  # type: ignore
+import packaging.version  # type: ignore
+from astropy.time import Time  # type: ignore
```
WilliamJamieson, line 22:
```suggestion
import asdf
import numpy as np
import packaging.version
from astropy.time import Time 
```
Left over from prototyping.

src/roman_datamodels/datamodels.py
```diff
@@ -33,7 +33,7 @@
             category=asdf.exceptions.AsdfDeprecationWarning,
             message=r"AsdfInFits has been deprecated.*",
         )
-        from asdf.fits_embed import AsdfInFits
+        from asdf.fits_embed import AsdfInFits  # type: ignore
```
WilliamJamieson, line 36:
```suggestion
        from asdf.fits_embed import AsdfInFits
```
Same

src/roman_datamodels/extensions.py
```diff
@@ -1,4 +1,4 @@
-from asdf.extension import ManifestExtension
+from asdf.extension import ManifestExtension  # type: ignore
```
WilliamJamieson, line 1:
```suggestion
from asdf.extension import ManifestExtension
```
Same

src/roman_datamodels/maker_utils.py
```diff
@@ -343,7 +343,7 @@ def mk_cal_step():
     -------
     roman_datamodels.stnode.CalStep
     """
-    calstep = stnode.CalStep()
+    calstep = stnode.CalStepValues()
```
WilliamJamieson, line 346:
```suggestion
    calstep = stnode.CalStep()
```
VSCode got a bit too aggressive here with its variable rename tool I used in other parts of the code.

src/roman_datamodels/random_utils.py
```diff
@@ -227,3 +227,38 @@ def generate_optical_element(seed=None):
...
+        group = base_list[index]
+        if group == 0:
+            continue
+
+        if group not in read_pattern:
+            read_pattern[group] = []
+
+        read_pattern[group].append(index)
+
+    return [read_pattern[key] for key in sorted(read_pattern.keys())]
```
WilliamJamieson, line 264: This is separate, but there was not a random `read_pattern` generator... So I wrote one in an early iteration of these changes. I think this should be a separate PR.
```suggestion
```

src/roman_datamodels/stnode.py
```diff
@@ -37,6 +39,20 @@
...
+def registration_off():
+    """
+    Context manager to turn off registration of classes.
+    """
+
+    global META_REGISTER
+    META_REGISTER = False
+    yield
+    META_REGISTER = True
+
```
WilliamJamieson, line 55: Enables temporary turning off registration of tagged object subclasses, so that the type stubs can be loaded as a module without breaking anything.

src/roman_datamodels/stnode.py
```diff
@@ -110,7 +126,7 @@ class DNode(UserDict):
     def __init__(self, node=None, parent=None, name=None):
         if node is None:
             self.__dict__["_data"] = {}
-        elif isinstance(node, dict):
+        elif isinstance(node, Mapping):
```
WilliamJamieson, line 129:
```suggestion
        elif isinstance(node, dict):
```
I think this can be reverted.

src/roman_datamodels/stnode.py
```diff
@@ -224,7 +240,7 @@ class LNode(UserList):
     def __init__(self, node=None):
         if node is None:
             self.data = []
-        elif isinstance(node, list):
+        elif isinstance(node, Iterable):
```
WilliamJamieson, line 243:
```suggestion
        elif isinstance(node, list):
```
I think  this can also be reverted.

src/roman_datamodels/stnode.py
```diff
@@ -404,7 +423,7 @@ def select_tag(self, obj, tags, ctx):
         return obj.tag
 
     def to_yaml_tree(self, obj, tag, ctx):
-        return obj._data
+        return dict(obj._data)
```
WilliamJamieson, line 426:
```suggestion
        return obj._data
```
I think this also can be reverted.

src/roman_datamodels/_typing/maker.py
```diff
@@ -0,0 +1,370 @@
...
+    1: (4096,),
+    2: (4096, 4096),
+    3: (2, 4096, 4096),
+    4: (2, 4, 4096, 4096),
+    5: (2, 8, 16, 32, 32),
+}
+
+REPLACE_NAME = {
+    "pass_": "pass",
+}
```
WilliamJamieson, line 73: This exists because I cannot use `pass` as a variable name in the way I define the type stub. It is a reserved keyword for Python and bad things happen if I do `pass: foo` as a variable declaration. So this fixes that.

pyproject.toml
```diff
@@ -28,7 +29,7 @@ test = [
     'pytest >=6.0.0',
     'pytest-doctestplus',
     'pytest-doctestplus >=0.10.0',
-    'pytest-env >= 0.8'
+    'pytest-env >= 0.8',
```
WilliamJamieson, line 32:
```suggestion
    'pytest-env >= 0.8'
```

.github/workflows/ci.yml
```diff
@@ -26,20 +26,20 @@ jobs:
...
           python-version: 3.11
-        - linux: test-cov-xdist
+        - linux: test-cov
           coverage: codecov
   test_upstream:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
     with:
       envs: |
-        - linux: test-rad-xdist
+        - linux: test-rad
```
WilliamJamieson, line 45: Dropping `xdist` testing because with the additional tests in this PR there are two issues with `xdist` in github actions.

1. `xdist` is causing memory size collisions. In the sense that by default action runners have 7Gb of memory shared across the two cores. GitHub will kill the action if it attempts to use more than this amount of memory at once. Some of the existing tests (prior to this PR) do use over 4Gb (some of the tests in this PR use over 4Gb too). It is possible for changes in the test ordering to cause two of these "largish" memory tests to hit at the same time, resulting in the action being killed. This reordering is inevitable given that the tests are dispatched on a first-come basis to the two cores. So adding/reordering tests may in cause this to occur. This happens under some conditions with these additional tests mixed in.
2. Also `xdist` can cause the CI to take longer than necessary:

	- https://github.com/spacetelescope/roman_datamodels/actions/runs/5006108889/jobs/8970943733 is a run with xdist used (7min).
	- https://github.com/spacetelescope/roman_datamodels/actions/runs/5006147741/jobs/8971028814 is without xdist (4min).

Also, note that without `xdist` there is little difference between lengths of test runs:

- https://github.com/spacetelescope/roman_datamodels/actions/runs/5006371967/jobs/8971569618?pr=181 (the head of this PR's changes, without `xdist`) takes 4ish min.
- https://github.com/spacetelescope/roman_datamodels/actions/runs/5006325192/jobs/8971454680 (head of main, has `xdist`) takes 3.5ish min.

Given the additional tests in this PR this indicates that `xdist` is not really doing anything for the CI.

This is not to say that `xdist` won't help make them faster in some circumstances, or make the tests always fail due to memory issues. The failure is GitHub specific, and the negligible change in run time is down to overhead vs core scaling. Locally, and/or use on different action runners (like the Large ones or an internally hosted one) will likely benefit from `xdist` because of the greater number of cores available and larger memory limits.

So I believe this is a reasonable change under the circumstances.

src/roman_datamodels/stnode.pyi
```diff
@@ -0,0 +1,543 @@
...
+        asn_pool: str
+        asn_type: str
+        asn_rule: str
+        version_id: str
+        code_version: str
+        degraded_status: str
+        program: int
+        target: int
+        constraints: str
+        products: stnode.LNode[_AssociationProduct]
```
PaulHuwe, line 543: I presume this would need upkeep as well?

src/roman_datamodels/_typing/base.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+import sys
+from typing import Generic, TypeVar
+
+import astropy.units as u
+import numpy as np
+
+if sys.version_info < (3, 10):
+    raise ImportError("Python 3.10+ is required to use the typing module features")
```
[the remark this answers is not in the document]
WilliamJamieson: Yes this fail as soon as it is imported under 3.9 because it is using syntax that is not present in python 3.9. That error is less clear than us raising an error at this point.

Since this module would only be imported during testing or followed by outside code tools I see no issue with not supporting python 3.9 here. Since the completion engines would simply not be able to generate completions.

src/roman_datamodels/_typing/maker.py
```diff
@@ -0,0 +1,373 @@
...
+import inspect
+import string
+import sys
+import warnings
+from pathlib import Path
+from typing import get_args, get_type_hints
+
+if sys.version_info < (3, 10):
+    raise ImportError("Python 3.10+ is required to use the typing module features")
+
```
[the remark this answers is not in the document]
WilliamJamieson: See https://github.com/spacetelescope/roman_datamodels/pull/181#discussion_r1198067851

src/roman_datamodels/_typing/maker.py
```diff
@@ -0,0 +1,373 @@
...
+    if issubclass(args[0], u.Quantity):
+        return u.Quantity(create_float(value, random), unit=args[1])
+
+    if args[0].__name__ in ENUM_CLASSES_BY_NAME:
+        if random:
+            components = [create_enum(args[0]) for _ in range(RNG.integers(low=LIST_0, high=LIST_1))]
+
+            return f"{args[1].join(components)}{args[1]}"
+
+        return "WFI_IMAGE|WFI_GRISM|WFI_PRISM|"
```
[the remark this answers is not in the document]
WilliamJamieson: This is no different than magic like https://github.com/spacetelescope/roman_datamodels/blob/dc212d8dd79cbca6f6e93e33c03e22e94e54c588/src/roman_datamodels/maker_utils.py#L682.

If deterministically creatable instances of these objects is in fact desirable, then stuff like this must be done. I can do it via a fixed seed if you would prefer it not be explicit.

src/roman_datamodels/stnode.pyi
WilliamJamieson: We already essentially do the same work with `factories.py` and `maker_utils.py`, I specifically built the constructor here to be able to fill it out with random data (like `factories.py`) or fixed data (`maker_utils.py`). This way I could replace both modules with construction from a singular, verified source. The duplication of effort between those two modules should have never occurred, instead both modes should have been consolidated.

This presumes of course, that we drop Python 3.9 support. Which I would argue is not a bad idea. According to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) support for Python 3.9 will be dropped on April 5, 2024. We will have to drop Python 3.9 support at that point because both NumPy and astropy will drop support for it then. This is well before any timeline which would actually operationally use the Roman pipeline in any meaningful fashion. Indeed, I would argue that we only support Python 3.11 since it is the only currently available version of Python which we will be still supported by NumPy/astropy by 2026 (Its support will be dropped April 24, 2026).

WilliamJamieson: Note that any replacement for `factories.py` and/or `maker_utils.py` by such a centralized constructor would simply map the interfaces in those two modules back to the interfaces provided by the centralized constructor. So for all intents and purposes, this type of change would not effect any outside code due to interface changes.

src/roman_datamodels/_typing/enums.py
```diff
@@ -0,0 +1,208 @@
...
+class SaturationRefType(StrEnum):
+    SATURATION = "SATURATION"
+
+
+class SuperbiasRefType(StrEnum):
+    BIAS = "BIAS"
+
+
+class WfiPhotomRefType(StrEnum):
+    PHOTOM = "PHOTOM"
```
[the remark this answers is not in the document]
WilliamJamieson: This module is entirely supported by Python 3.9, so I see no reason not to do such refactoring sooner rather than later.  Assuming this change is accepted. However, I feel such changes are outside the scope of the proposed changes.

src/roman_datamodels/stnode.pyi
PaulHuwe: Okay - I was just asking to gauge what future support would entail.

PaulHuwe (approved):
I like this and think it is a good change, but I want @ddavis-stsci to approve before merging.

.github/workflows/ci.yml
zacharyburnett: This sounds entirely reasonable to me. Should we also remove `-xdist` in `ci_cron.yaml` as well?

ddavis-stsci (changes requested):
The documentation should also be updated to describe these updates and the stnode use in general. Examples and a list of generated types would also be valuable.

PR #182, opened 2023-05-16
Update the pytest `addopts` configuration

WilliamJamieson:
`roman_datamodels` currently does not capture standard out when running its tests. This was configured off for some reason in the pytest configuration and it makes it annoying for debugging simple issues via print statements.

This PR removes that configuration and adds `--color=yes` so the actions output is colored (if possible) and the `--doctest-rst` flag to actually enable doc-tests (though we currently do not have any, but do have the doctest plugin installed).

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ minversion = 4.6
 doctest_plus = true
 doctest_rst = true
 text_file_format = 'rst'
-addopts = '--show-capture=no'
+addopts = '--color=yes --doctest-rst'
 testpaths = ['tests']
 filterwarnings = [
     # This error replaces pytest-openfiles


ddavis-stsci (dismissed):
Global changes to the behavior of the software, such as this, need to be discussed with the group before updates are made.

pllim (approved):
This brings the setting in line with the widely used and widely tested `astropy`, so LGTM. Thanks!

https://github.com/astropy/astropy/blob/a429c3984a14c995584455e51a6f3d7d9c16e914/setup.cfg#L128

nden:
This change does not introduce anything new to the dependency list. It simply enables testing in examples in the documentation. And we can discuss it on this PR, no need for a separate meeting imho.

WilliamJamieson:
> Global changes to the behavior of the software, such as this, need to be discussed with the group before updates are made.

Discussion is the purpose of opening PRs.

I fail to see what you are objecting to, these are minor changes to the `pytest` configuration:

1. `--show-capture=no` is a legacy of when @perrygreenfield spilt `roman_datamodels` from `stdatamodels`, see https://github.com/spacetelescope/roman_datamodels/commit/403f6b55622f71bd93afa69c85838735f441c030 specifically: https://github.com/spacetelescope/roman_datamodels/blob/403f6b55622f71bd93afa69c85838735f441c030/setup.cfg#L65 An offline discussion with @perrygreenfield indicates that this was the result of copy/paste, not some specific reason.

	- Note `--show-capture` controls whether or not `pytest` displays the `stdout` it captures from a test when it fails as part of its failing message. This is often useful for debugging to see `stdout` logging messages and things like `print` statement returns. Turning it off (its on by default) means one has to force `pytest` to return all `stdout` in its main run output which separates it from the error message. This makes it rather annoying to match things between the two.

2. `--color=yes` simply tells `pytest` to "colorize" its main output if the terminal session running it supports colors. Basically, passing tests are marked with green symbols, failing tests are marked with red symbols, and warnings issued result in yellow/orange symbols. The `%` mark will remain green if all tests up to the point it was created were passing, yellow/orange if everything up-to that point is passing but a warning occurred, and red if any failure has occurred. This is quite useful when looking at github actions runs as they are running (such as when they take a long time) to spot if a failure has occurred in CI. This is useful when debugging environments you don't have easy local access to.

3. `--doctest-rst` tells `pytest` to search `rst` files for possible doctests and add them to its test list. This is necessary for the `pytest-doctestplus` plugin to do anything useful. Otherwise, it will get no tests from our documentation to generate tests from. Granted, we currently do not have any doctests in our documentation, but we do actually have this as a test dependency and configuration for its behavior, so we should probably enable it so it does pick up doctests if/when we add them.

perrygreenfield:
I agree completely with William.

nden (approved):

nden (approved):

PR #191, opened 2023-05-30
Remove `STUserDict`

WilliamJamieson:
`STUserDict` is essentially a copy of the Python standard library's `collections.UserDict` source code with the change of name and the change of the `data` (`UserDict`) variable to `_data` (`STUserDict`). This copy was a little while ago and is no longer "in-sync" with the standard library's version.

It turns out that this approach is a bit "overkill" for what `DNode` (what uses `STUserDict`) actually needs. Instead it just needs to be an abstract mutable mapping (abc.MutableMapping).

This PR switches `DNode` from using `STUserDict` to `abc.MutableMapping` which solves this issue. It also removes `STUserDict` because it is no-longer necessary.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/298/


WilliamJamieson:
Leaving this as a draft until all the regression tests are functional again. This will need to be carefully tested.

docs/conf.py
```diff
@@ -26,25 +26,11 @@
...
-    return find_mod_objs(args[0], onlylocals=True)
-
-
-def patch_automodapi(app):
-    """Monkey-patch the automodapi extension to exclude imported members"""
-    from sphinx_automodapi import automodsumm
-
-    automodsumm.find_mod_objs = find_mod_objs_patched
-
-
```
WilliamJamieson, line 41: Monkey patches like this are generally a bad idea. The whole reason for the need of this patch was poor construction of `stuserdict.py` module to begin with. https://github.com/spacetelescope/roman_datamodels/pull/156#issuecomment-1520536988 details a couple of different ways of properly resolving this without monkey patching. However, all of those issues are moot because this PR removes `stuserdict.py`.

WilliamJamieson:
The regression tests passed: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/298/ so this is ready.

PaulHuwe (approved):
LGTM

PR #192, opened 2023-05-31
Update ModelContainer to comply with stpipe's ABC pseudo-protocol.

mairanteodoro:
Resolves:
- [RCAL-568](https://jira.stsci.edu/browse/RCAL-568);
- [AL-743](https://jira.stsci.edu/browse/AL-743).

This PR addresses updates to `roman_datamodels.datamodels.ModelContainer` in order to make it compliant with `stpipe`'s `AbstractDataModel` pseudo-protocol.

**Checklist**
- [X] Added entry in `CHANGES.rst` under the corresponding subsection
- [ ] updated relevant tests

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+0.15.1 (unreleased)
+===================
+
+- ModelContainer is now compliant with stpipe's AbstractDataModel. [#192]
+
 0.15.0 (2023-05-15)
 ===================
 
--- a/src/roman_datamodels/datamodels.py
+++ b/src/roman_datamodels/datamodels.py
@@ -519,6 +519,29 @@ class ModelContainer(Iterable):
     def to_association(self):
         pass
 
+    @property
+    def crds_observatory(self):
+        """
+        Get the CRDS observatory for this container.  Used when selecting
+        step/pipeline parameter files when the container is a pipeline input.
+
+        Returns
+        -------
+        str
+        """
+        return DataModel().crds_observatory
+
+    @property
+    def get_crds_parameters(self):
+        """
+        Get parameters used by CRDS to select references for this model.
+
+        Returns
+        -------
+        dict
+        """
+        return DataModel().get_crds_parameters()
+
 
 class FlatRefModel(DataModel):
     pass


WilliamJamieson (commented):
These changes look fine. Assuming they solve the issues experienced in `romancal`.

There is not a good way to "test" that this is correct without installing `stpipe` as a testing dependency. If we do want to verify the satisfaction of the protocol, then we simply need to import `AbstractDatamodel` from `stpipe` and perform an `isinstance` check.

@PaulHuwe and @ddavis-stsci do either of you object to adding stpipe as a test dependency?

mairanteodoro:
For now, I used the regression test that we're implementing for `TweakReg` as a test for the code implemented in this PR. The regression test was failing before because it was not an instance of `AbstractDataModel`, which led us to create [AL-743](https://jira.stsci.edu/browse/AL-743). With the changes in this PR, the regression test for `TweakReg` is passing.

mairanteodoro:
@ddavis-stsci Four months ago, we (@nden, @ddavis-stsci, and @stscieisenhamer) agreed to implement `ModelContainer` without inheriting `DataModel`. The specific discussion about this can be found on Slack: https://stsci.slack.com/archives/C01FP900P38/p1674139315067459

WilliamJamieson:
> I still don't really understand this, if I just define
> 
> ```
> class ModelContainer(rdm.DataModel, Iterable):
> ...     pass
> ```
> 
> I seem to get what you are trying to include in the same way as the the other modules.
> 
> ```
> >>> dir(ModelContainer)
> ['__abstractmethods__',..., 'copy', 
> 'crds_observatory', 
> 'get_crds_parameters', 'get_primary_array_name', ... 'validate']
> ```
> 
> Am I missing some issue?

Note that there is no need. The check in `stpipe` is not strictly based on inheritance (`datamodel` does not inherit from `stpipe`'s `AbstractDatamodel`) instead it is a check that an object satisfies a given interface.

Since `ModelContainer` was not supposed to inherit from `Datamodel` (for other reasons). This is a good solution as it is the one employed by `Datamodel` also. 

If we really want to implement this via inheritance, we really should instead factor the given interface into an abstracted base class or mixin instead. That will maintain the function separation between `Datamodel` and `ModelContainer` (`Datamodel` has or will likely accumulate stuff that is functionally irrelevant or counter to the assumptions of the `ModelContainer` object).

PR #190, opened 2023-05-30
RCAL-484: Update datamodels for reference pixel correction

WilliamJamieson:
Resolves [RCAL-484](https://jira.stsci.edu/browse/RCAL-484)

This is the matching PR to spacetelescope/rad#270, in that it implements the updates to `roman_datamodels` to support adding the reference pixel correction's reference file.

**Checklist**
- [ ] Added entry in `CHANGES.rst` under the corresponding subsection
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/XXX/

PR #193, opened 2023-05-31
Refactor the `maker_utils`

WilliamJamieson:
This https://github.com/spacetelescope/roman_datamodels/pull/190#pullrequestreview-1451797922 indicates that there is an expectation that a `maker_util` both exists and is "properly tested. If such an expectation exists, then there should be tests to enforce those expectations, such as the tests for the factories.

This pull request adds those tests and refactors the maker_utils into an organized submodule to both facilitate making the code easier to read and easier to test.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster with the changes in spacetelescope/romancal#712. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/221/


src/roman_datamodels/maker_utils/__init__.py
```diff
@@ -0,0 +1,28 @@
+from ._basic_meta import *  # noqa: F403
+from ._common_meta import *  # noqa: F403
+from ._datamodels import *  # noqa: F403
+from ._ref_files import *  # noqa: F403
```
WilliamJamieson, line 4: This recreates the API from the original single module, the `*` makes it so that nothing is missed. We can add `__all__` definitions to the contained modules to control this.

src/roman_datamodels/maker_utils/_datamodels.py
```diff
@@ -0,0 +1,432 @@
...
+        af.tree = {"roman": rampfitoutput}
+        af.write_to(filepath)
+    else:
+        return rampfitoutput
+
+
+def mk_rampfitoutput(**kwargs):
+    warnings.warn("mk_rampfitoutput is deprecated. Use mk_rampfit_output instead.", DeprecationWarning)
+
+    return mk_ramp_fit_output(**kwargs)
```
WilliamJamieson, line 321: This method was a deviation from the naming convention which had a clear path to a proper name. I have left it in for now to not break anything, but it raises a `DeprecationWarning` now.

tests/test_maker_utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+    schema_keys = set()
+    subschemas = [schema]
+    if "allOf" in schema:
+        subschemas.extend(schema["allOf"])
+    for subschema in subschemas:
+        schema_keys.update(subschema.get("properties", {}).keys())
+
+    diff = instance_keys - schema_keys
+    assert len(diff) == 0, "Dummy instance has extra keys: " + ", ".join(diff)
```
WilliamJamieson, line 44: These are basically the same thing as the factory tests use, adapted for the maker utilities

tests/test_maker_utils.py
```diff
@@ -0,0 +1,78 @@
...
+def test_make_ref_tests(util):
+    """
+    Meta test to confirm that correct tests exist for each maker utility.
+    """
+    from . import test_models as tests
+
+    name = util[3:]
+
+    assert hasattr(tests, f"test_make_{name}")
+    assert hasattr(tests, f"test_opening_{name}_ref")
```
WilliamJamieson, line 68: These tests ensure that the tests for the maker utility function alluded to here: https://github.com/spacetelescope/roman_datamodels/pull/190#pullrequestreview-1451797922 are implemented.
The first makes sure the `__all__` for the second covers every reference file. The second checks that the aforementioned tests are actually present.

tests/__init__.py
WilliamJamieson, line 1: This is needed so I can import a test module in order to test if that module has all the desired tests.

src/roman_datamodels/maker_utils/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+# These makers have special names to reflect the nature of their use in the pipeline
+SPECIAL_MAKERS = {
+    "WfiScienceRaw": "mk_level1_science_raw",
+    "WfiImage": "mk_level2_image",
+    "WfiMosaic": "mk_level3_mosaic",
+}
```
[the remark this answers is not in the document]
PaulHuwe: Eventually, I think we should, but yes - there is a cost to changing these names downstream.

PaulHuwe (changes requested):
LGTM, please add regression tests.

PaulHuwe (approved):
LGTM

PR #197, opened 2023-06-02
Remove the factories

### LINKED ISSUE #196: Deprecate and/or remove `testing.factories`
During the last Roman developers tag-up (June 2, 2023), PR #193 was discussed. 

As part of the discussion it was pointed out that #193's changes (mostly in adding through automatic testing) enable the `maker_utils` to construct objects equivalent to those constructed by the `testing.factories`.  Meaning that both sets of construction functions largely duplicate each others function (generating a valid `stnode` object). Thus it was proposed that we entirely remove the `testing.factories` because the `maker_utils` have proven to be more useful. A less drastic option would be to first deprecate `testing.factories` and then remove it.

Currently, `romancal` does not use any of the `testing.factories` methods in its actual code, it only mentions them in its documentation, which would be relatively easy to update to mention `maker_utils` instead. The `testing.factories` are used quite extensively for testing `roman_datamodels`. However, since #193 adds tests to ensure that the `maker_utils` stay in sync with RAD in exactly the same way as the current tests ensuring `testing.factories` is, it is a rather simple process to transition all the tests relying on `testing.factories` to using `maker_utils`.

The primary difference is `maker_utils` mostly fills the `stnode` objects with fixed so called "dummy" data (e.g. data arrays of all zeros, and `NOSTR` for strings), while `testing.factories` attempts to fill all items with randomly generated data.

Note that there are a few places where `maker_utils` is non-deterministic (it makes random choices):

- https://github.com/spacetelescope/roman_datamodels/blob/d61648727d4d88d6e2bb2c35b74a04e5ad578c6e/src/roman_datamodels/maker_utils.py#L252
- https://github.com/spacetelescope/roman_datamodels/blob/d61648727d4d88d6e2bb2c35b74a04e5ad578c6e/src/roman_datamodels/maker_utils.py#L1243
- https://github.com/spacetelescope/roman_datamodels/blob/d61648727d4d88d6e2bb2c35b74a04e5ad578c6e/src/roman_datamodels/maker_utils.py#L1183

For consistency with the rest of `maker_utils` it might be advantageous to update these to use deterministic (fixed) choices.

This issue is intended to be a place for the Roman developers to discuss removing `testing.factories` or other alternatives.

WilliamJamieson:
Closes #196

This PR is a demonstration for #196 of removing the `testing.factories` and replacing its usage with `maker_utils`.

Note that this is based on PR #193 which refactors the `maker_utils` so that they are semantically equivalent to the `testing.factories`.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/309/


WilliamJamieson:
Note that a small tweak to the `romancal` documentation is needed, if we merge this. This is simply changing:

- https://github.com/spacetelescope/romancal/blob/00fb99e9ddceb6c055cf8d71e6273569f088d5e7/docs/roman/datamodels/metadata.rst?plain=1#L19-L21 
- https://github.com/spacetelescope/romancal/blob/00fb99e9ddceb6c055cf8d71e6273569f088d5e7/docs/roman/datamodels/models.rst?plain=1#L53C1-L58

from using `testing.factories` to `maker_utils`, will be required.

PaulHuwe (approved):
LGTM

WilliamJamieson:
I have no idea why this is suddenly creating 143 errors.

ddavis-stsci (changes requested):
The issue seems to be that you are making default size images for the unit tests? 
These should be smaller test images that should not strain the memory and run much faster. See the unit tests for ramp_fitting in rcal.

WilliamJamieson:
@ddavis-stsci this is all for parameterized tests so it is not that simple to just specify non-default sizes as not all the datamodels and stnode objects have the same parameterization.

WilliamJamieson:
> > @ddavis-stsci this is all for parameterized tests so it is not that simple to just specify non-default sizes as not all the datamodels and stnode objects have the same parameterization.
> 
> In this case I'm not sure that this is ready to merge without more design work.

This is a limitation of the `maker_utils` design not the changes in this pull request. These changes are outside the scope of using the factories, which are not used except in the tests within `roman_datamodels` themselves.

WilliamJamieson:
Converting to a draft because @ddavis-stsci's requested changes are outside the scope of this PR. So I have opened #207 implementing them. I will be rebasing this PR on those changes.

ddavis-stsci (approved):
= 1966 passed in 58.81s =

PR #198, opened 2023-06-02
Add `datamodel` maker

WilliamJamieson:
When developing tests in `romancal` it was a little annoying to construct the `datamodel` you needed, as the `maker_utils`/`testing.factories` all construct the `stnode` objects, not the actual `datamodels` themselves.

This PR provides a convenience function (`maker_utils.mk_datamodel`) to do this for you. For any `datamodel` `type` (in the `MODEL_REGISTRY`) it automatically grabs the correct `maker_util` and then uses the it to construct data which it then feeds into the `datamodel` class to construct a full fledged `datamodel` of the desired type.

For example if one wants a `RampModel` the following will construct a `RampModel` filled with dummy data:

```python
from roman_datamodels.datamodels import RampModel
from roman_datamodels.maker_utils import mk_datamodel

model = mk_datamodel(RampModel)
```

Note that this is based on #193 and so it will remain a draft for now.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/310/

### DIFF
--- a/src/roman_datamodels/maker_utils/__init__.py
+++ b/src/roman_datamodels/maker_utils/__init__.py
@@ -1,3 +1,5 @@
+from roman_datamodels.datamodels import MODEL_REGISTRY as _MODEL_REGISTRY  # Hide from public API
+
 from ._basic_meta import *  # noqa: F403
 from ._common_meta import *  # noqa: F403
 from ._datamodels import *  # noqa: F403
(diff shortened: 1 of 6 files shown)


src/roman_datamodels/maker_utils/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    model_class : type
+        One of the datamodel subclasses from datamodel
+    **kwargs
+        Additional or overridden attributes.
+
+    Returns
+    -------
+    `roman_datamodels.datamodels.Datamodel`
+    """
+    return model_class(mk_node(NODE_REGISTRY[model_class], **kwargs))
```
WilliamJamieson, line 67: This is the actual magic of this whole PR. This automatically pick up any new `Datamodels` added to the `MODEL_REGISTRY` be able to start working with no additional developer input outside of providing `maker_utils` for the `stnode` structure, which the other tests require them to do.

PaulHuwe (approved):
LGTM - like this addition

PR #200, opened 2023-06-05
Bugfix: Datamodels can be initialized with incorrect node.

WilliamJamieson:
@nden noticed that she could initialize any `Datamodel` with any `stnode`, without any issues. This can lead to many unexpected issues, as the `Datamodel` objects are wrappers around the `stnode`. The `stnode` is what handles all the validation and serialization of the `Datamodel` not the `Datamodel` itself. Indeed, if one initializes a `Datamodel` with the wrong `stnode` then when the model is deserialized, one will not get the same datamodel back.

This PR fixes this issue by crosschecking the registered top-level `stnode` for each `Datamodel` against the `stnode` passed to it during construction, an error is raised if the `stnode` is not of the correct type. 

**Checklist**
- [ ] updated relevant tests


tests/test_models.py
```diff
@@ -814,11 +816,26 @@ def test_model_validate_without_save():
...
+@pytest.mark.parametrize("node", datamodels.model_registry.keys())
+@pytest.mark.parametrize("correct, model", datamodels.model_registry.items())
+def test_model_only_init_with_correct_node(node, correct, model):
+    """
+    Datamodels should only be initializable with the correct node in the model_registry.
+
+    This checks that it can be initiallized with the correct node, and that it cannot be
+    with any other node.
+    """
+    img = create_node(node)
```
WilliamJamieson, line 839: This test is rather slow because the `testing.factories` constructors can be extremely slow due to lots of random number generation. #193 can make this much more efficient because it provides a `maker_util` that can determine which node to make, that is only available to `testing.factories` at this time.

src/roman_datamodels/datamodels.py
```diff
@@ -85,6 +85,9 @@ def __init__(self, init=None, **kwargs):
             self._asdf = asdffile
             self._instance = asdffile.tree["roman"]
         elif isinstance(init, stnode.TaggedObjectNode):
+            if not isinstance(self, model_registry.get(init.__class__)):
+                expected = {mdl: node for node, mdl in model_registry.items()}[self.__class__].__name__
+                raise ValueError(f"TaggedObjectNode: {init.__class__.__name__} is not of the type expected. Expected {expected}")
```
WilliamJamieson, line 90: Should this be a `ValidationError` instead.

PR #201, opened 2023-06-06
Refactor the `datamodels` module into a sub-package.

WilliamJamieson:
Currently, the `datamodels` module (`datamodels.py`) is getting a bit complex and is beginning to trip over itself. For example, the need for and discussion in #176 makes clear that at minimum the `open` function needs to be factored into its own separate module. Moreover, there is a lot of cross purposes in the module which make it both long and difficult to follow.

This PR attempts to rectify these issues by turning `datamodels` into a sub-package (directory with `__init__.py`). By doing this we can split the various functions of `datamodels.py` into their own modules. The split is the following:

- `DataModel` is moved into the `datamodels._core` module.
	-  `DataModel` really is just an "abstract" core class which while containing the majority of the relevant datamodel code, is distinct from all the subclasses using it.
	- This means organizationally, it makes sense to push it into its own module away from everything else.
- `open` is moved into the `datamodels._utils` module.
	- This solves the issue "solved" by #176, in a "cleaner" way by fully moving and renaming it into its own module.
	- Additionally, its complete separation from the datamodels, cleans up concerns by making so the "working" datamodels can be distinctly separated from any other code.
	- Finally, the completely separated code means it can be renamed away from shadowing the builtin `open` function, but this still enables us to exactly maintain the current datamodels API via re-naming it in the `datamodels.__init__`. This means `from roman_datamodels.datamodels import open` or `from roman_datamodels import datamodels as rdm` with `rdm.open` both still fully work.
- ~`ModelContainer` is moved into the `datamodels._container` module.~ (removed by #204)
	- ~`ModelContainer` is not actually a "datamodel" in the sense that it is a subclass of `DataModel`, instead it distinctly a container of `DataModel` objects.~
	- ~This warrants a separation into a module wholly responsible for this which is not mixed with any of the datamodel definitions.~
- All the remaining datamodels are moved into the `datamodels._datamodels` module.
	- Having all these in their own file will make it simpler to add new models as there is clearly one place for them.
	- Basically, this is what we really wanted by stuffing everything in `datamodels.py` originally.

Other improvements gained by this refactor include:

- Forcing datamodels to specify their `node_type`, which explicitly makes the association between a datamodel and its top-level node explicit.
- Automatic creation of the `model_registry`, the datamodels now have enough in-built information to add themselves to the model registry upon their creation (as a class).

Note this supersedes and so it can close #176.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/242/


PaulHuwe (approved):
LGTM minus one suggestion

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class IpcRefModel(_DataModel):
+    _node_type = stnode.IpcRef
+
+
+class LinearityRefModel(_DataModel):
+    _node_type = stnode.LinearityRef
+
+    def get_primary_array_name(self):
```
perrygreenfield, line 133: This seems a little misleading, since this is an override rather than implying that it may be overridedn. Same for the following cases  where this method is defined.

perrygreenfield (commented):
Looks good except for what I think are misleading docstrings in a few of the datamodel subclasses.

src/roman_datamodels/datamodels/_datamodels.py
WilliamJamieson: I didn't add those docstrings that's what was there before I made my changes.

PR #208, opened 2023-06-16
measure coverage on package not tests

braingram:
It appears the coverage measurement in the CI is measuring coverage of the tests, not the package as seen by the coverage results:
https://app.codecov.io/gh/spacetelescope/roman_datamodels
and an example of a recent coverage run:
https://github.com/spacetelescope/roman_datamodels/actions/runs/5283947700/jobs/9560855551#step:10:97

Is this expected or should the coverage be run against the package?

This PR switches the coverage measurement to run against the package and not the tests. It might be preferable to include both the package and tests (this can be helpful to catch test definitions with the same name and other bugs in the tests) but I'm not sure how to do this. @zacharyburnett is there an incantation that will enable coverage for both?

**Checklist**
- [ ] Added entry in `CHANGES.rst` under the corresponding subsection
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/XXX/

PR #209, opened 2023-06-16
Add test for DataModel.schema_uri and update to non-deprecated usage

braingram:
`TagDefinition.schema_uri` is deprecated since asdf (since 2.9)
https://asdf.readthedocs.io/en/stable/api/asdf.extension.TagDefinition.html#asdf.extension.TagDefinition.schema_uri
and is currently used by `DataModel.schema_uri`.

This PR updates `DataModel.schema_uri` to use the non-deprecated `TagDefinition.scham_uris` (note the `s`). This PR also adds a test to cover `DataModel.schema_uri` which was not previously being tests (see https://github.com/spacetelescope/roman_datamodels/pull/208).

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/234/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,9 @@
 - Update ``roman_datamodels`` to support the new reference file for the
   reference pixel correction. [#190]
 
+- Update ``DataModel.schema_uri`` to use non-deprecated
+  ``TagDefinition.schema_uris`` from asdf [#209]
+
 0.15.0 (2023-05-15)
 ===================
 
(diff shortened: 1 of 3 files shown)


WilliamJamieson (changes requested):
Please add a change log entry and link to a passing run of the regression tests.

WilliamJamieson (approved):
I am surprised that this was missed, it looks like the relevant test was not doing what we expected. Thanks for catching it.

PR #213, opened 2023-06-21
Move stnode into a sub-package

WilliamJamieson:
This PR moves `stnode` into a sub-package and then splits module apart into separate modules each handling one part of the functionality of `stnode`. In addition to the split, I have adjusted the names of some of the private functions and added further documentation of stnode. Hopefully, the split of responsibilities combined with the additional inline code documentation will make `stnode` easier to maintain.

In addition, to the split/document, I also simplified a few parts of the code:

1. I removed the `metaclass` based registry of node classes, and replaced it with `__init_subclass__` registry. The metaclasses are more than we need for this application.
2. I made all STNode classes dynamically generated (solved scalar/list node corner case).
3. As part of the dynamic changes, I changed the code to using a mixin based system. That is for nodes that we need to define additional code on, instead of making a direct subclass, instead I changed it to a mixin which is automatically detected and mixed into the class dynamicall.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/241/


perrygreenfield (commented):
Other than a minor docstring error, it looks good. I had to refresh my brain by looking at the original code (even though I wrote much of it, but not the dynamic aspects) and that took some time.

src/roman_datamodels/stnode/_tagged.py
```diff
@@ -0,0 +1,152 @@
...
+
+def get_schema_from_tag(ctx, tag):
+    """
+    Look up and load ASDF's schema corresponding to the tag_uri.
+
+    Parameters
+    ----------
+    ctx :
+        An ASDF file context.
+    tag_uri : str
```
WilliamJamieson, line 33:
```suggestion
    tag : str
```

perrygreenfield (commented):
LGTM

perrygreenfield (approved):
LGTM

PR #214, opened 2023-06-22
Make use of the new `datamodel_name` keyword in RAD

WilliamJamieson:
This is a follow on PR to spacetelescope/rad#278, which describes the reasoning behind the `datamodel_name` and implements it.

This PR simply adds the consistency testing described by spacetelescope/rad#278 to `roman_datamodels`.

Note that this PR is built on top of #201 as it directly associates a datamodel type with an stnode type which is created directly from the RAD schemas. It is possible to do this without #201, but those changes make it much simpler to implement.

**Checklist**
- [x] updated relevant tests
- [x] ~Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/XXX/~ This only effects roman_datamodels testing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,9 @@
 ===================
 - Remove the ``random_utils`` module and make ``maker_utils`` entirely deterministic. [#217]
 
+- Add tests to ensure consistency between file-level schemas in RAD and the corresponding
+  datamodels in ``roman_datamodels``. [#214]
+
 0.16.1 (2023-06-27)
 ===================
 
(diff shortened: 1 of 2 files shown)


WilliamJamieson:
Note that it is possible to take this concept one step further because now their exists enough information in RAD to dynamically create the datamodel objects themselves from RAD (though some of the additional functionality for various models would need to be mixed in to these objects as what was done by #213).

I am not suggesting that we should dynamically generate the datamodel objects as then the datamodels would run into the same set of issues which are described and partially solved by #181. Instead I am simply pointing out the possibility, which the Roman developers should discuss in more detail.

PaulHuwe (approved):
LGTM

PaulHuwe (approved):
LGTM

PR #217, opened 2023-06-22
Remove the `random_utils`

### LINKED ISSUE #215: Should we remove randomness from the `maker_utils`?
It is my understanding that one of the main differences between the `maker_utils` and the `factories` was that the `maker_utils` were supposed to deterministically generate the data to fill a datamodel. However, there are three exceptions to this:

1. https://github.com/spacetelescope/roman_datamodels/blob/a271364d530e35c1b3580fdc414e417f44620c53/src/roman_datamodels/maker_utils/_common_meta.py#L210
2. https://github.com/spacetelescope/roman_datamodels/blob/a271364d530e35c1b3580fdc414e417f44620c53/src/roman_datamodels/maker_utils/_common_meta.py#L490
3. https://github.com/spacetelescope/roman_datamodels/blob/a271364d530e35c1b3580fdc414e417f44620c53/src/roman_datamodels/maker_utils/_datamodels.py#L370

All three of these are fairly straight forward to resolve. Thus should we simply replace this randomness with something deterministic?

### LINKED ISSUE #216: Should we drop the `random_utils` module?
The main use for `random_utils` was the `factories` module, which was removed by #197 in favor of just using the `maker_utils`.

Note that outside of the now removed `factories` (and the tests for the `random_utils`) there is very little in the way of any actual code using the `random_utils`. Indeed, there are no direct uses of the `random_utils` outside of `roman_datamodels` itself (i.e. they are not used anywhere in `romancal` at this time). There are two exceptions, both in the `maker_utils` ([imported here](https://github.com/spacetelescope/roman_datamodels/blob/a271364d530e35c1b3580fdc414e417f44620c53/src/roman_datamodels/maker_utils/_common_meta.py#L5)):

1. https://github.com/spacetelescope/roman_datamodels/blob/a271364d530e35c1b3580fdc414e417f44620c53/src/roman_datamodels/maker_utils/_common_meta.py#L210
2. https://github.com/spacetelescope/roman_datamodels/blob/a271364d530e35c1b3580fdc414e417f44620c53/src/roman_datamodels/maker_utils/_common_meta.py#L490

While outside the scope of this issue, #215 proposes removing these do make the `maker_utils` fully deterministic; meaning that, if #215 is pursued there will be no uses of `random_utils` at all meaning we should just drop `random_utils`.

If we decide against #215, we only actually need two functions from all the `random_utils` so I would instead propose moving those functions inside `maker_utils` as part of the private API and then remove `random_utils`.

WilliamJamieson:
Closes #215 and closes #216

This PR removes all randomness from the `maker_utils` and then removes the `random_utls` as they are no-longer needed.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/324/


PaulHuwe (approved):
LGTM

PR #218, opened 2023-06-22
Update maker_utils so that saved nodes are still returned

WilliamJamieson:
@stscieisenhamer requested that the `maker_utils` return the nodes even when they are saving the node to ASDF. This will not effect any current uses as the output will be thrown away, but it will enable one to use the node made by the maker_util and saved to a file without having to open that file.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/326/ (all are passing except the tests noted in https://github.com/spacetelescope/stcal/pull/173#issuecomment-1601669335 due to a recent stcal update)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 - Add tests to ensure consistency between file-level schemas in RAD and the corresponding
   datamodels in ``roman_datamodels``. [#214]
 
+- Make ``maker_utils`` return the node when writing the node to a file. [#218]
+
 0.16.1 (2023-06-27)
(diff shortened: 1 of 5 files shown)


stscieisenhamer (approved):
LGTM.

Note: I never completed the jira issue filing. So, nevermind on that.

PaulHuwe (approved):
LGTM *sweep sweep sweep*

PR #220, opened 2023-06-22
Refactor the ASDF extension into stnode

WilliamJamieson:
This PR refactors the ASDF extension so that it exists entirely within the stnode sub-package because it is really an ASDF extension to handle serialization/deserialization of the stnode objects. This also makes the API for the extension explicitly private, as it should only be used by `roman_datamodels` to register itself as an ASDF extension.

**Checklist**
- [x] updated relevant tests


WilliamJamieson:
This PR is based off the changes in #213, so it will remain a draft until that PR is merged.

src/roman_datamodels/stnode/_converters.py
```diff
@@ -1,19 +1,38 @@
...
 from astropy.time import Time
 
-from ._registry import LIST_NODE_CLASSES_BY_TAG, OBJECT_NODE_CLASSES_BY_TAG, SCALAR_NODE_CLASSES_BY_TAG
+from ._registry import LIST_NODE_CLASSES_BY_TAG, NODE_CONVERTERS, OBJECT_NODE_CLASSES_BY_TAG, SCALAR_NODE_CLASSES_BY_TAG
 
 __all__ = [
     "TaggedObjectNodeConverter",
     "TaggedListNodeConverter",
     "TaggedScalarNodeConverter",
+    "NODE_EXTENSIONS",
```
braingram, line 13: Am I understanding this correctly that `NODE_EXTENSIONS` added here will be exposed at `roman_datamodels.stnode.NODE_EXTENSIONS`? This seems consistent with `NODE_CLASSES` which has a comment mentioning it is public: https://github.com/spacetelescope/roman_datamodels/blob/2dd4d7f41b0eb9074f8e32e6bee525c47881e3d2/src/roman_datamodels/stnode/_stnode.py#L47-L53
However I don't believe either of these appear in the docs:
https://roman-datamodels--220.org.readthedocs.build/en/220/roman_datamodels/datamodels/developer_api.html#module-roman_datamodels.stnode
Should these be in the docs?

braingram (approved):
One comment/question about `NODE_EXTENSIONS` and docs but looks good to me!

PR #221, opened 2023-06-23
Proof of Concept: Dynamically Generated DataModels

WilliamJamieson:
After the roman tag-up this week (22 June 2023), I was left wondering how difficult would it be to dynamically generate the datamodels from RAD now that the `datamodel_name` meta data has been added to the schemas.

As it turns out, this is not that difficult if one builds on the proposed changes in PRs #213 and #214 (This PR is based off a a rebase of #214 onto #213). This PR move the datamodels from being statically defined in code to dynamic code generation as discussed.

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/src/roman_datamodels/datamodels/_factory.py
+++ b/src/roman_datamodels/datamodels/_factory.py
@@ -1,0 +1,28 @@
+from roman_datamodels import stnode
+
+from . import _mixins
+from ._core import DataModel
+
+
+def datamodel_names():
+    for tag in stnode._stnode.DATAMODELS_MANIFEST["tags"]:
+        schema = stnode._factories.load_schema_from_uri(tag["schema_uri"])
+        if "datamodel_name" in schema:
+            yield stnode.OBJECT_NODE_CLASSES_BY_TAG[tag["tag_uri"]], schema["datamodel_name"]
+
+
+def datamodel_factory(node_type, datamodel_name):
+    if hasattr(_mixins, mixin := f"{datamodel_name}Mixin"):
+        class_type = (DataModel, getattr(_mixins, mixin))
+    else:
+        class_type = (DataModel,)
+
+    return type(
+        datamodel_name,
+        class_type,
+        {
+            "_node_type": node_type,
+            "__module__": "roman_datamodels.datamodels",
+            "__doc__": f"Roman {datamodel_name} model",
+        },
+    )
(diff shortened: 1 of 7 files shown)


src/roman_datamodels/datamodels/_factory.py
```diff
@@ -0,0 +1,55 @@
...
+        The base STNode class to use as the base for the DataModel
+    datamodel_name : str
+        The name of the DataModel to create
+
+    Returns
+    -------
+    A DataModel object class
+    """
+    if hasattr(_mixins, mixin := f"{datamodel_name}Mixin"):
+        class_type = (DataModel, getattr(_mixins, mixin))
```
WilliamJamieson, line 43: There is a bug here, things like the 
`get_primary_array_name` cannot be properly overridden as desired.

This is because the "order" of inheritance in Python is right to left meaning it applies the right most class structure first then proceeds to the left. Thus for the `get_primary_array_name` example the Mixin will load its version and then it will be overloaded by the DataModel object. This is the reverse of what we want.

This should not effect anything added by the Mixin (as one would normally do), but since this is the way I am dynamically injecting alterations to dynamically created objects, we need to pay attention to this ordering.

src/roman_datamodels/datamodels/_mixins.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class LinearityRefModelMixin:
+    """
+    Mixin class for dynamically generated LinearityRefModel
+    """
+
+    def get_primary_array_name(self):
+        """
+        Returns the name "primary" array for this model, which
```
perrygreenfield, line 77: This should be changed to reflect that it is the override, rather than intended to be overridden.

src/roman_datamodels/datamodels/_mixins.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class MaskRefModelMixin:
+    """
+    Mixin class for dynamically generated MaskRefModel
+    """
+
+    def get_primary_array_name(self):
+        """
+        Returns the name "primary" array for this model, which
```
perrygreenfield, line 100: Likewise.

perrygreenfield (commented):
Looks good to me and pretty straightforward, aside from the needed changes to a couple docstrings.

Issue #150, opened 2023-04-13
Strange `ResourceWarning`


WilliamJamieson:
Background: in PR #142, the use of `pytest-openfiles` was switched out for catching `ResourceWarnings` directly with `pytest`. This should work because Python will issue a `ResourceWarning` if a file handle is left open when the file goes out of scope (that is leaving a file open). For the most part this will work as a drop in replacement; however, under certain circumstances `pytest` is unable to raise the `ResourceWarning` as an error, resulting in a `PytestUnraisableExceptionWarning`. Typically, these are raised when testing error messages using the `pytest.raises` context manger. The interesting thing is that these `PytestUnraisableExceptionWarnings` are being issued for things not caught by `pytest-openfiles`. For example, in #143 some new warnings show up in the [CI](https://github.com/spacetelescope/roman_datamodels/actions/runs/4659291564/jobs/8245992937?pr=143#step:10:70):
```
tests/test_filetype.py::test_filetype
  /home/runner/work/roman_datamodels/roman_datamodels/.tox/test-xdist/lib/python3.11/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
  
  Traceback (most recent call last):
    File "/home/runner/work/roman_datamodels/roman_datamodels/tests/test_filetype.py", line 19, in test_filetype
      file_8 = filetype.check(open(DATA_DIRECTORY / "fake.asdf"))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/runner/work/roman_datamodels/roman_datamodels/tests/data/fake.asdf' mode='r' encoding='UTF-8'>
  
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))

tests/test_models.py::test_make_saturation
  /home/runner/work/roman_datamodels/roman_datamodels/.tox/test-xdist/lib/python3.11/site-packages/_pytest/unraisableexception.py:[78](https://github.com/spacetelescope/roman_datamodels/actions/runs/4659291564/jobs/8245992937?pr=143#step:10:79): PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
  
  Traceback (most recent call last):
[...]

PR #232, opened 2023-07-05
Do not copy datamodels when opening an already open datamodel.

schlafly:
This PR starts to try to reduce memory usage in romancal by changing the default behavior of rdm.open(datamodel) to just return the existing datamodel rather than to make a copy.  This likely has important downstream consequences that I haven't fully sussed out, though!


WilliamJamieson:
We should be a bit careful here, currently this fails https://github.com/spacetelescope/roman_datamodels/blob/ce423ee7a2cf44f421581f02bad4349c795026da/tests/test_open.py#L98-L100

The note is absolutely correct, in that if we don't do the copy being removed, then calling `close` on some upstream reference to the model object here will cause downstream problems because then inadvertently one can end up trying to access information from a closed model. Indeed, this behavior stretches all the way back to PR #1.

Currently, we are a bit caviler about sometimes calling close and sometimes not calling `close` on models. Especially since `rdm_open` is used extensively as a context manager. This cavalierness may cause a couple of issues for us:

1. We accidentally close a model which was "opened" and then modified by a step when the step exits. This is east to cause when using `rdm_open` as a context manager. More generally this can be caused via directly calling `close` somewhere on the model (which will be annoying to debug).
2. We leave the backing files to a model open and loose reference to those files due to loosing reference to the model, which can cause all sorts of subtle problems. In theory, the `__del__` method on the model should clean up the open files for us as it closes the model when the reference count drops to zero, but as the python [documentation](https://docs.python.org/3/reference/datamodel.html#object.__del__) for that method indicates `__del__` is not guaranteed to always be called; moreover, the  [data model documentation
](https://docs.python.org/3/reference/datamodel.html#objects-values-and-types) states:

>Objects are never explicitly destroyed; however, when they become unreachable they may be garbage-collected. An implementation is allowed to postpone garbage collection or omit it altogether — it is a matter of implementation quality how garbage collection is implemented, as long as no objects are collected that are still reachable.

In effect this means we cannot guarantee that our models will be properly cleaned up by `__del__`. Indeed, that is what was causing #150 (see #231 fix). 

These two issues are both solved by the copy being removed here because it disconnects models "opened" during a given step from models passed in. So we both never accidentally close a model we need open, and can properly close the files for models when needed.

I personally agree with the changes here as it is silly to copy things for no real reason. However, we need to approach this very carefully or otherwise we might cause ourselves very difficult to solve bugs.

schlafly:
Yes.  The way I think of this, if we open a file, we're responsible for cleaning it up.  If we accept an object, the person who made the object is responsible for cleaning it up (or we wait for it to be garbage collected).

The current pipeline practice of using
`with rdm.open(in) as data`
as a generic "open a file or ~do nothing" seems problematic to me.  In the former case we often want to clean up files.  In the latter case we don't really want to do anything, but we don't want to clean up the object, since it was sent to us by someone else.  We currently have a uniform interface by making a copy and then cleaning that copy up (unless opening with the target argument, which I doubt we handle well?), so that the caller's copy isn't cleaned up.  But that's an unacceptable overhead here.

Maybe a better solution would be for `rdm.open(in)` to return a clone of `in`.  That looks like it's by default a shallow copy that will ~turn off context handling and might be exactly what we want?

schlafly:
I think this is ready for review.  This changes the behavior of rdm.open(...) to return a shallow rather than full copy.  rdm.open(...) is used throughout romancal and stcal to open already opened data models, and the previous behavior of making copies for this purpose wasted a lot of memory.  This has some implications in terms of what we're allowed to return from the pipeline, which leads to me making lazy_loading the default in the romancal pipeline.  That's discussed in more detail in the related romancal PR here: https://github.com/spacetelescope/romancal/pull/774

PaulHuwe:
Can you add a plwishmaster link (presumably with the RCAL-774 ticket) to show this passing regression tests?

schlafly:
romancal regression tests with this PR here: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/281/
I don't think I ran separate datamodels regression tests and would have to look at this again next week.

schlafly:
(and note that I don't remember but that I don't think regression tests on romancal would pass with _only_ this PR.  One needs the other work on romancal too.)

ddavis-stsci:
I'm confused, what other romancal work does the memory work depend on?
Are you saying that the steps don't run with these changes?

ddavis-stsci:
That is fine. I just misread your reply to Paul. Sorry for the confusion.

PaulHuwe (approved):
LGTM

schlafly:
Merged in updates from main, reran regression tests; merging.
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/285/

PR #59, opened 2022-01-12
RCAL-193: Update Mask Code in Ramp, Flat, and Jump step and tests to RDM

PaulHuwe:
Updated rampfit and flat maker utilities to support the same functionality as the other model maker functions. 
Streamlined and commented all maker utility functions. 
Added tests to complete coverage of roman_datamodels/testing/utils.py
Cleaned out some deprecated code.

Issue #51, opened 2021-11-04
Data model constructors should return object instance used in constructor if right type


perrygreenfield:
Currently if one supplies an existing instance of a datamodel (say RampModel) to one of the datamodel constructors (say RampModel), and the instance is the same type as the constructor, the constructor should return that instance. Currently it raises an error.

This may require some trickery to effectively replace self with the instance (I will have to brush up on this)

nden:
Fixed in #52

PR #235, opened 2023-07-06
Remove the `target` keyword from `rdm_open`

WilliamJamieson:
This PR addresses issue #51 and the partial solution to it in PR #52. The solution in #52 was to add the `target` keyword to `rdm_open`, however it did not fix the actual issue from #51. Then in PR #59, the tests for the solution in #52 were completely removed, and after that point the `target` argument completely ceased being used in `roman_datamodels` and `romancal`.

The issue at hand in #51 is that if we pass a data model instance to the datamodel constructor (`<data_model_class>(*args, **kwargs)`), then we don't want it to raise an error. The discussion and solution in #52 additionally indicates the we should pass the exact object out with no alterations rather than constructing a new object.

This pull request adds a test reproducing the error described in #51 together with the additional constraints from #52. It then solves the issue via injecting a new `__new__` method for the `DataModel` objects which detects and then handles this case special case. It then reverts the partial solution from #52. 

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/267/

### DIFF
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -56,7 +56,22 @@ class DataModel(abc.ABC):
         # Add to registry
         MODEL_REGISTRY[cls._node_type] = cls
 
+    def __new__(cls, init=None, **kwargs):
+        """
+        Handle the case where one passes in an already instantiated version
+        of the model. In this case the constructor should just directly return
+        the model.
+        """
+        if init.__class__.__name__ == cls.__name__:
+            return init
+
+        return super().__new__(cls)
+
     def __init__(self, init=None, **kwargs):
+        if isinstance(init, self.__class__):
+            # Due to __new__ above, this is already initialized.
+            return
+
         self._iscopy = False
         self._shape = None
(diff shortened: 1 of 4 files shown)


src/roman_datamodels/datamodels/_core.py
```diff
@@ -56,7 +56,22 @@ def __init_subclass__(cls, **kwargs):
         # Add to registry
         MODEL_REGISTRY[cls._node_type] = cls
 
+    def __new__(cls, init=None, **kwargs):
+        """
+        Handle the case where one passes in an already instantiated version
+        of the model. In this case the constructor should just directly return
+        the model.
+        """
+        if init.__class__.__name__ == cls.__name__:
```
WilliamJamieson, line 65: Bad things happen (infinite recursion) if one tries to `isinstance` checks here.

PaulHuwe (approved):
LGTM

PR #204, opened 2023-06-08
RCAL-568: remove ModelContainer from datamodels.

mairanteodoro:
Resolves [RCAL-568](https://jira.stsci.edu/browse/RCAL-568)

This PR removes `ModelContainer` from `roman_datamodels.datamodels`.

**Checklist**
- [X] Added entry in `CHANGES.rst` under the corresponding subsection
- [X] updated relevant tests
- [X] updated relevant documentation
- [ ] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/XXX/

PR #143, opened 2023-04-07
Fix all the test warnings

WilliamJamieson:
Fixes #129

PR #238, opened 2023-07-12
Make tests covered by codecov

### LINKED ISSUE #210: Run coverage on test code in addition to the package code
Running coverage on the code that makes up the test suite can help to catch bugs in the tests including:
- tests that share the same name (coverage will show that only one of these tests was run and pytest will not complain when this happens)
- never used conditional statements (here is one [innocuous example from this package](https://app.codecov.io/gh/spacetelescope/roman_datamodels/blob/main/tests%2Ftest_factories.py#L41))
- empty parameterization of tests

WilliamJamieson:
This PR makes the tests covered by codecov so that we can detect when test code is not being executed.

Fixes #210

**Checklist**
- [x] updated relevant tests
- [ ] ~Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/XXX/~ Only touches testing configurations.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,40 +2,9 @@ import asdf
 import pytest
 import yaml
 
-from roman_datamodels import datamodels
-from roman_datamodels import maker_utils as utils
-
 MANIFEST = yaml.safe_load(asdf.get_config().resource_manager["asdf://stsci.edu/datamodels/roman/manifests/datamodels-1.0"])
 
 
 @pytest.fixture(scope="session")
 def manifest():
     return MANIFEST
-
-
-@pytest.fixture(name="set_up_list_of_l2_files")
-def set_up_list_of_l2_files(tmp_path, request):
-    # generate a list of n filepaths and files to be read later on by ModelContainer
-    marker = request.node.get_closest_marker("set_up_list_of_l2_files_data")
-    number_of_files_to_create = marker.args[0]
-    type_of_returned_object = marker.args[1]
-
-    result_list = []
-    for i in range(number_of_files_to_create):
-        filepath = tmp_path / f"test_model_container_input_as_list_of_filepaths_{i:02}.asdf"
-        # create L2 file using filepath
-        utils.mk_level2_image(filepath=filepath)
-        if type_of_returned_object == "asdf":
-            # append filepath to filepath list
-            result_list.append(str(filepath))
-        elif type_of_returned_object == "datamodel":
-            # parse ASDF file as RDM
-            datamodel = datamodels.open(str(filepath))
-            # append datamodel to datamodel list
-            result_list.append(datamodel)
-
-    return result_list
-
-
-def pytest_configure(config):
-    config.addinivalue_line("markers", "set_up_list_of_l2_files_data(number_of_files_to_create, type_of_returned_object): ")
--- a/tests/test_maker_utils.py
+++ b/tests/test_maker_utils.py
@@ -136,7 +136,6 @@ def test_override_data(node_class):
         Generate a mutated value for a given value.
             Note:
                 - Time is a special case because it's constructor is picky.
-                - Pure lists need to be preserved.
                 - TaggedScalarNodes need their type preserved.
         """
         if isinstance(value, Time):
--- a/tests/test_maker_utils.py
+++ b/tests/test_maker_utils.py
@@ -47,6 +47,8 @@ def test_no_extra_fields(node_class, manifest):
 
     schema_keys = set()
     subschemas = [schema]
+    if "allOf" in schema:
+        subschemas.extend(schema["allOf"])  # pragma: no cover
     for subschema in subschemas:
         schema_keys.update(subschema.get("properties", {}).keys())
 
--- a/tests/test_maker_utils.py
+++ b/tests/test_maker_utils.py
@@ -47,8 +47,6 @@ def test_no_extra_fields(node_class, manifest):
 
     schema_keys = set()
     subschemas = [schema]
-    if "allOf" in schema:
-        subschemas.extend(schema["allOf"])
     for subschema in subschemas:
         schema_keys.update(subschema.get("properties", {}).keys())
 
@@ -143,9 +141,6 @@ def test_override_data(node_class):
         """
         if isinstance(value, Time):
             return value + 1 * u.day
-
-        if isinstance(value, list):
-            return [mock.MagicMock()]
 
         if isinstance(value, stnode.TaggedScalarNode):
             return value.__class__(mutate_value(value.__class__.__bases__[0](value)))
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
     with:
       envs: |
         - linux: py39-oldestdeps-cov
+          coverage: codecov
         - linux: py39
         - linux: py310
         - linux: py311
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@
 
 - Clean up the ``rdm_open`` function. [#233]
 
+- Include tests in coverage and turn testing warnings into errors. [#238]
+
 0.16.1 (2023-06-27)
 ===================
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,9 +69,8 @@ text_file_format = 'rst'
 addopts = '--color=yes --doctest-rst'
 testpaths = ['tests']
 filterwarnings = [
-    # This error replaces pytest-openfiles
-    "error::ResourceWarning",
-    "error::pytest.PytestUnraisableExceptionWarning",
+    "error",
+    "ignore:numpy.ndarray size changed:RuntimeWarning",
 ]
 env = [
     "ROMAN_VALIDATE=true",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,16 +78,6 @@ env = [
     "ROMAN_STRICT_VALIDATION=true",
 ]
 
-[tool.coverage.run]
-omit = [
-    'roman_datamodels/pyproject.toml',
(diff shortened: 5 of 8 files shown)


tests/conftest.py
```diff
@@ -2,40 +2,9 @@
...
-
 MANIFEST = yaml.safe_load(asdf.get_config().resource_manager["asdf://stsci.edu/datamodels/roman/manifests/datamodels-1.0"])
 
 
 @pytest.fixture(scope="session")
 def manifest():
     return MANIFEST
-
-
-@pytest.fixture(name="set_up_list_of_l2_files")
```
braingram, line 16: Are the two functions here unused? Looking back at the history they were added in #143 but I'm not seeing where they were used.

tests/test_maker_utils.py
```diff
@@ -144,9 +142,6 @@ def mutate_value(value):
         if isinstance(value, Time):
             return value + 1 * u.day
 
-        if isinstance(value, list):
```
braingram, line 147: The docstring mentions that 'Pure lists need to be preserved'

https://github.com/spacetelescope/roman_datamodels/blob/e9d93fdef8c80f26036bd09a0404d9ba89a157fd/tests/test_maker_utils.py#L136-L148

Is this no longer necessary? I'm not familiar with this code so would you describe why it's no longer needed and if so, update the docstring?

tests/test_maker_utils.py
```diff
@@ -47,8 +47,6 @@ def test_no_extra_fields(node_class, manifest):
 
     schema_keys = set()
     subschemas = [schema]
-    if "allOf" in schema:
-        subschemas.extend(schema["allOf"])
```
braingram, line 51: I could use help understanding what this does. Would you explain why this isn't needed? I'm thinking it would cover the case where a RAD schema extends an existing schema by doing something like:
```yaml
allOf:
  - $ref: other_schema-1.0.0.yaml
  - extra_stuff..
```
kind of like what is done in [file_date-1.0.0.yaml](https://github.com/spacetelescope/rad/blob/1c7dd7fee74ba7a8b57c6250ed4399e35199a7a8/src/rad/resources/schemas/tagged_scalars/file_date-1.0.0.yaml#L8-L9). I'm not seeing any schema for the nodes tested here that use that pattern. Is there something in the design of the members of NODE_CLASSES that prevents this kind of schema extension (and makes this part of the test unnecessary in addition to being unused)?

braingram (changes requested):
Thanks for the changes and adding the tests to the coverage measurement.

I have a few questions about the changes to the tests and would like to hear if the other maintainers have any comments about turning all warnings into errors in the test suite (that seems like an improvement to me).

tests/test_maker_utils.py
WilliamJamieson: It was never needed.

tests/conftest.py
WilliamJamieson: They were totally ununsed.

WilliamJamieson: The back history is that this was used by, but never added by the `ModelContainer` which was removed by PR #204. Clearly #204 missed removing this. The #143 PR added it so that the `ModelContainer` tests would actually be run correctly (`pytest` was raising a clear warning about this).

zacharyburnett (approved):
looks like coverage is working and reporting, so this LGTM

tests/test_maker_utils.py
WilliamJamieson: Look at the parameterization: https://github.com/spacetelescope/roman_datamodels/blob/e9d93fdef8c80f26036bd09a0404d9ba89a157fd/tests/test_maker_utils.py#L38

The `FileDate` object generated by the `file_date-1.0.0.yaml` is NOT a `TaggedObjectNode` it is a `TaggedScalarNode`.

This test in not applicable to `TaggedScalarNode` objects because by construction they cannot have any `keys`. The same applies to `TaggedScalarNode` objects as well.

WilliamJamieson: In general, the `TaggedObjectNode` objects should not need to have a top-level `allOf` combiner.

braingram: Thanks for the extra details.

I provided the link to file date to provide an example of the pattern. Thanks for the confirmation that as I mentioned the pattern is not currently used in rad schemas that turn into TaggedObjectNodes.

If I add a schema that uses the pattern to rad and a maker utility to roman_datamodels all tests pass without issue if the changes are based off of rad main and roman_datamodels main. If however I base the changes off of the source branch for this PR (with the lines discussed here removed) the roman_datamodels (but not rad) tests fail with:
```
AssertionError: Dummy instance has extra keys: ...
```
due to the `test_no_extra_fields` test failing to step into the allOf at the base of the added schema.

The pattern of extending one schema with an allOf and ref is used extensively in JWST schemas and there appears to be nothing in RAD that prevents this pattern and nothing in roman_datamodels that prevents turning an extended schema into functional TaggedObjectNode.

If this pattern should not be used in roman wouldn't it be preferable to either:
- add a restriction in the rad schema (if possible)
- add a test in rad (several appear to exist to further check schemas)

instead of removing these lines so that a schema that extends another schema fails the `test_no_extra_fields` test?

If so, those changes seem outside the scope of this PR.

WilliamJamieson: > The pattern of extending one schema with an allOf and ref is used extensively in JWST schemas and there appears to be nothing in RAD that prevents this pattern and nothing in roman_datamodels that prevents turning an extended schema into functional TaggedObjectNode.
> 
> If this pattern should not be used in roman wouldn't it be preferable to either:
> 
> * add a restriction in the rad schema (if possible)
> * add a test in rad (several appear to exist to further check schemas)
> 
> instead of removing these lines so that a schema that extends another schema fails the `test_no_extra_fields` test?
> 
> If so, those changes seem outside the scope of this PR.

As it stands, virtually all changes to the structure of a schema in RAD will naturally induce failures in `roman_datamodels` because currently there is not a solution to dynamically generating node objects filled with null data from the schemas. Effectively, this means any change such as you describe is expected to induce failures in `roman_datamodels`. Thus it is expected that all changes in RAD are developed alongside changes to `roman_datamodels`.

If there becomes a case where an `allOf` combiner is needed in a `TaggedObjectNode`'s schema then this can be re-introduced. However, there is not currently a use case for this and it actively discourages unnecessary usage of a top level `allOf` combiner. I have already had to clean up several unnecessary uses of the `allOf` combiner in RAD, for some examples see spacetelescope/rad#267. While these unnecessary uses of the `allOf` do not effect how the schema is applied, the do unnecessarily complicate the schemas, which makes it more difficult for new/external users to understand the structure of their datamodels (the structure is dynamically generated from these schemas meaning the only reference users have for the structure is the schemas, which is already a barrier to some users).

I would rather not have to continually notice and fix small issues like spacetelescope/rad#267, but it is involved to add such a test in RAD. This is a relatively simple thing to address here, and as already stated there is an implied expectation of co-developing most changes significant changes between RAD and `roman_datamodels` meaning the home of such checks does not matter.

All this being said I will revoke this change as there are clear objections to it.

braingram: Thanks for the clarification. I don't have any objection to preventing the extension of existing schemas (but I don't also have as much involvement in this project as @ddavis-stsci @PaulHuwe @schlafly etc).

Thanks for helping me understand the impact of removing these lines of code. I don't think this change should be in a PR changing the coverage measurement to include tests.

WilliamJamieson:
> This is not universally a good idea. 

I strongly disagree with this statement in the context of test suites. It is generally considered best practice to turn warnings into errors within test suites. There are several reasons for this including:

1. Warnings are generally signals to downstream developers that there is or will be a potential issue.
2. It is quite easy to ignore even rather serious warnings that really ought to be addressed, see https://github.com/spacetelescope/romancal/actions/runs/5506145209/jobs/10034448151#step:10:170. 
3. Test suites are intended to discover issues and potential issues with code prior to it being released.

That is we should have to explicitly address (and think about) any issues or potential issues our test suite.

Organizations such as `astropy` have learned to do this "the hard way" by having to deal with the fallout of not directly addressing the issues flagged by warnings until they start causing real problems. Resolving this fallout required a significant effort from several developers, which could have instead been addressed piecemeal without sudden significant efforts.

> We can get warnings from the time package when leap second information is not available. 

The EFRA warnings about leap seconds we were seeing were the result of the fact that the times we were choosing where too far in the future for there to exist a leap second table to cover those times. As the reason for this is well understood, it is entirely reasonable to filter these warnings out. This does not prevent us from ignoring or catching warnings, it simply turns unhandled warnings into errors so that we are forced to document/track or address them up front.

>We can also get deprecation warnings while we have a version pinned and waiting for our dependencies to upgrade their packages.

Deprecation warnings are the main reason for why warnings in test suites should be errors. They exist to signal to downstream code (such as our code) that something will break in the future. As such we should be forced to address and document these issues. Since warnings do not cause failures, the CI cannot directly signal this issue, which then requires someone to notice and flag these issues.

PaulHuwe:
I also agree with changing warnings to errors for tests, with the proviso that we can specify certain warnings to ignore (e.g. leap seconds).

WilliamJamieson:
> I also agree with changing warnings to errors for tests, with the proviso that we can specify certain warnings to ignore (e.g. leap seconds).

Yes we can absolutely do that, for example `astropy` ignores several: https://github.com/astropy/astropy/blob/98b399d52d730d91c2cd29babe1e8545caff219c/setup.cfg#L133-L154. 

As far as the leap second warnings, those were resolved via removing the `factories`. The `maker_utils` pick a time default which does not raise the leap second warning. However, we can add this warning to be ignored if it comes up again in addition to any other warnings we actively decide that we can ignore.

braingram (approved):
Thanks! LGTM

PR #239, opened 2023-07-14
Implement sp-repo-review suggestions

WilliamJamieson:
The Scientific Python Organization has published a [development guide](https://learn.scientific-python.org/development/). As part of this guide they have built the [`sp-repo-review`](https://github.com/scientific-python/repo-review) tool which automatically checks for a package following the development guide. The majority of these recommendations create improvements in the code. This PR implements the majority (everything except using `mypy`) of the suggestions from `sp-repo-review` tool.

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,14 +1,18 @@
 <!-- If this PR closes a JIRA ticket, make sure the title starts with the JIRA issue number,
 for example RCAL-1234: <Fix a bug> -->
+
 Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/RCAL-nnnn)
 
 <!-- If this PR closes a GitHub issue, reference it here by its number -->
+
 Closes #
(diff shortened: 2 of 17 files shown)


ddavis-stsci:
Is this or will this be part of the SCSB coding standards? I don't see it on the Code Development page.

WilliamJamieson:
This is an implementation of the Scientific Python organization's development guide for packages. SCSB can choose to adopt all or part of it or not, this was simply done as part of an interaction session with Scientific Python. As users and contributors to the general scientific Python community I believe SCSB should at least consider the guidelines which that community has formally discussed and agreed upon. Indeed, that community is actively interested in continuing to maintain such its guide and is open to suggestions for it. 

In my opinion, evaluating whether or not we choose to adopt all or parts of this guide is best done via actively applying it to part of our code base in order observe and discuss the results. `roman_datamodels` happens to be both a relatively small package and one that I am familiar enough with to quickly handle the edge cases that could have arisen, making a good candidate for me to use as part of a short informational session.

I believe the changes made represent positive or neutral improvements. Moreover, the resulting code improvements are all enforceable via automated tools (applied by our existing pre-commit system) which in most cases can automatically apply fixes for the issues they catch.

If anyone disagrees with the changes made, please list them here. I would like to have an open discussion of any concerns about these suggestions.

ddavis-stsci (changes requested):
It is fine using roman_datamodels as a test bed for the proposal. 
This should not be merged into the public branch until SCSB decides 
that we should be following these guidelines.

PR #251, opened 2023-07-26
Clean up some parts of `datamodels._core`

### LINKED ISSUE #247: Fix empty DataModel initialization
_Issue [RCAL-612](https://jira.stsci.edu/browse/RCAL-612) was created on JIRA by [William Jamieson](https://jira.stsci.edu/secure/ViewProfile.jspa?name=wjamieson):_

It has been reported that initializing an empty DataModel of any type results in the model not being initialized around a DNode instance. This results in the inability to directly add any data to the model without reaching into its private state.

The init for DataModel should be updated to handle this case.

WilliamJamieson:
Resolves [RCAL-612](https://jira.stsci.edu/browse/RCAL-612)

Closes #247

This PR cleans up parts of the `datamodels._core` module and adds the ability to properly initialize an "empty" data model, in so far as that if an empty model is created, one can now successfully add information to it. However, the empty model will not validate until it is fully flushed out, so the recommendation is still to use the appropriate `maker_util`.

**Checklist**
- [x] updated relevant tests
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/294/. The failure appears to be the intermittent ASDF file open failure we sometimes get when a the network gets a bit slow.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@
 - Update minimum version of numpy to 1.22 as this is the oldest version of numpy
   which is currently supported. [#258]
 
+- Fix the initialization of empty DataModels and clean up the datamodel core. [#251]
+
 0.17.1 (2023-08-03)
 ===================
 
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -77,6 +77,18 @@ class DataModel(abc.ABC):
         self._instance = None
         self._asdf = None
 
+        if isinstance(init, stnode.TaggedObjectNode):
+            if not isinstance(self, MODEL_REGISTRY.get(init.__class__)):
+                expected = {mdl: node for node, mdl in MODEL_REGISTRY.items()}[self.__class__].__name__
+                raise ValidationError(
+                    f"TaggedObjectNode: {init.__class__.__name__} is not of the type expected. Expected {expected}"
+                )
+            with validate.nuke_validation():
+                self._instance = init
+                self._asdf = asdf.AsdfFile({"roman": init})
+
+                return
+
         if init is None:
             self._asdf = self.open_asdf(init=None, **kwargs)
         elif isinstance(init, (str, bytes, PurePath)):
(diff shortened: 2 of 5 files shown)


src/roman_datamodels/datamodels/_utils.py
```diff
@@ -95,3 +95,6 @@ def rdm_open(init, memmap=False, **kwargs):
 
         asdf_file.close()
         raise TypeError(f"Unknown datamodel type: {model_type}")
+
+        asdf_file.close()
+        raise TypeError(f"Unknown datamodel type: {model_type}")
```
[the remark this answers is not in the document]
WilliamJamieson: Thats a rebase error

CHANGES.rst
```diff
@@ -15,6 +15,8 @@
 
 - Fix newly required units from rand [#256]
 
+- Fix the initialization of empty DataModels and clean up the datamodel core. [#251]
```
nden, line 18: This should be moved to the `unreleased` section

src/roman_datamodels/datamodels/_core.py
```diff
@@ -305,24 +295,30 @@ def get_crds_parameters(self):
...
-        return crds_header
 
     def validate(self):
         """
         Re-validate the model instance against the tags
         """
         validate.value_change(self._instance, pass_invalid_values=False, strict_validation=True)
 
+    @property
+    def asdf(self):
```
braingram, line 311: Making this public might give the wrong impression that the underlying asdf file can be accessed and modified as part of the `DataModel` interface. This does not appear to be the case, assuming we have a model `m`
```
m.asdf['my_info'] = 42
assert m.asdf['my_info'] == 42
m.save('foo.asdf')
with asdf.open('foo.asdf') as af:
    af['my_info']  # KeyError as 'my_info' was not saved
```
This property (or at least the handling of `_asdf is None` appears needed for `info` `search` and `schema_info` below. Are there options for keeping this private?

WilliamJamieson: This has now been addressed using a decorator.

braingram (approved):
Thanks for addressing the comment. LGTM.

PR #253, opened 2023-07-31
AL-751: Allow DNode and LNode instances to be assigned to tree attributes

perrygreenfield:
Previously, if one obtained a DNode or LNode subclassed object from referencing a tree attribute, and then subsequently assigned it to the same attribute, a ValidationError was triggered. This was very non-intuitive. This PR permits this to be done, at the cost of not doing immediate validation in such cases (postponing such validation until the model was written, or validation was explicitly requested). Doing so makes the changes much simpler. Support for immediate validation will require much more work. As a result, this implementation allows the assignment of inconsistent objects to the attribute. 

Closes #

This PR addresses ...

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/src/roman_datamodels/stnode/_node.py
+++ b/src/roman_datamodels/stnode/_node.py
@@ -151,9 +151,7 @@ class DNode(MutableMapping):
                 if not (isinstance(value, DNode) or isinstance(value, LNode)):
                     if will_validate():
                         schema = _get_schema_for_property(self._schema(), key)
-                        if schema == {} or _validate(key, value, schema, self.ctx):
-                            # self._data[key] = value
-                            pass
+                        schema == {} or _validate(key, value, schema, self.ctx)
                 self._data[key] = value
             else:
                 raise AttributeError(f"No such attribute ({key}) found in node")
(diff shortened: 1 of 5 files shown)


src/roman_datamodels/stnode/_node.py
```diff
@@ -142,16 +142,19 @@ def __getattr__(self, key):
...
-
-                    if schema == {} or _validate(key, value, schema, self.ctx):
-                        self._data[key] = value
-                self.__dict__["_data"][key] = value
+                if not (isinstance(value, DNode) or isinstance(value, LNode)):
+                    if will_validate():
+                        schema = _get_schema_for_property(self._schema(), key)
+                        if schema == {} or _validate(key, value, schema, self.ctx):
+                            # self._data[key] = value
+                            pass
```
WilliamJamieson, line 156:
```suggestion
                        schema == {} or _validate(key, value, schema, self.ctx)
```
The "if statement" serves no purpose. If `schema` is in fact an empty dictionary `_validate` will not be called, otherwise it will.

WilliamJamieson (changes requested):

WilliamJamieson (changes requested):

WilliamJamieson (approved):
These changes look good to me on the surface. However, the regression tests are currently non-functional due to changes in RAD which are required by `roman_datamodels` at this time.

I'll leave it up to @PaulHuwe  to determine if we are comfortable merging these prior to running the regression tests.

perrygreenfield:
Superseded by PR #284

PR #258, opened 2023-08-08
Bump min versions

WilliamJamieson:
The oldest versions pins for `roman_datamodels` are currently [broken](https://github.com/spacetelescope/roman_datamodels/actions/runs/5783174538/job/15671465965). This is due to a breaking change in unit representations introduced in astropy/astropy#14439, which first appeared in astropy 5.3.0. So the min astropy version has been bumped to 5.3.0.

This necessitated an additional bump in numpy version at least 1.21. However, numpy 1.22 is the oldest supported version by numpy, so it makes sense to bump to 1.22.

**Checklist**
- [x] updated relevant tests
- ~[ ] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/XXX/~ The regression tests use the latest versions of these dependencies on PyPi not the minimum versions. However, this may require version bumps in `romancal` for the oldest dependencies job to remain functional.


pllim:
This looks reasonable. 5.0.x is super old by now and getting dropped at the end of 2023.

nden (approved):

PR #266, opened 2023-08-30
Remove `system_packages` option from RTD

WilliamJamieson:
We were not using this option (we simply turned it of explicitly). RTD has now removed this option so we need to remove it as well.

**Checklist**
- [ ] updated relevant tests


PaulHuwe (approved):
LGTM

PR #281, opened 2023-10-11
Fix dependency to be consistent with romancal

WilliamJamieson:
Currently the exact branch of `rad` is specified by this project. Unfortunately, this conflicts with the specification for dev dependencies in `romancal`, that is `romancal` does not have an `@main` decoration. This means that when attempting to install this version of `roman_datamodels` using the dev dependencies for `romancal` an "version conflict" occurs; meaning that none of the regression tests will even install properly. 

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     'numpy >=1.22',
     'astropy >=5.3.0',
     # 'rad >=0.17.1',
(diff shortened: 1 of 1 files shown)


zacharyburnett (approved):
we may also have to remove `.git` to match `romancal`'s `requirements-dev-st.txt`, depending on how exact the resolver is

PaulHuwe:
Ah, thanks for the catch. LGTM

PR #284, opened 2023-10-18
RCAL-695 generalize attribute assignment and creation

### LINKED ISSUE #54: Object nodes should allow new attributes to be set with dot notation if present in schema
Currently, only attributes that have already been assigned can be set using `.foo = 'bar'`.  Otherwise, the attribute needs to be set using `['foo'] = 'bar'`.  For example, this doesn't work:

```python
node = stnode.WfiImage()
node.cal_logs = []
```
but this does:
```python
node = stnode.WfiImage()
node["cal_logs"] = []
node.cal_logs = ["Hello"]
```

It would be convenient if attributes described by the schema could always be set with the dot operator.

perrygreenfield:
Resolves [RCAL-695](https://jira.stsci.edu/browse/RCAL-695)

Closes #54

This PR addresses allows new node attributes to be created using dot notation so long as the attribute is consistent with the schema (e.g., an explicit property or matches a patternProperty).

It also enables validation of the created attribute's value, as well as validating object instances assigned to existing attributes.

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -641,8 +641,27 @@ def test_node_assignment():
     darkexp = darkmodel.meta.exposure
     assert isinstance(darkexp, stnode.DNode)
     darkexp.ngroups = darkexp.ngroups + 1
-    assert darkexp.ngroups == 7
+    assert(darkexp.ngroups == 7)
     darkmodel.meta.exposure = darkexp
+
+# Test that a currently undefined attribute can be assigned using dot notation
+# so long as the attribute is defined in the coresponding schema.
+def test_node_new_attribute_assignment():
+    exp = stnode.Exposure()
+    with pytest.raises(AttributeError):
+        exp.bozo = 0
+    exp.ngroups = 5
+    assert exp.ngroups == 5
+    # Test patternProperties attribute case
+    photmod = utils.mk_wfi_img_photom()
+    phottab = photmod.phot_table
+    newphottab = {'F062': phottab['F062']}
+    photmod.phot_table = newphottab
+    photmod.phot_table.F213 = phottab['F213']
+    with pytest.raises(AttributeError):
+        photmod.phot_table.F214 = phottab['F213']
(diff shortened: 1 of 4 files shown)


PaulHuwe:
Can you add a PLWishMaster link to passed RCAL regression tests with this?

braingram:
I ran the regression tests here:
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/429/pipeline/216
It passed with no failures.

PaulHuwe (approved):
LGTM

tests/test_models.py
```diff
@@ -645,6 +645,26 @@ def test_node_assignment():
     darkmodel.meta.exposure = darkexp
 
 
+# Test that a currently undefined attribute can be assigned using dot notation
+# so long as the attribute is defined in the corresponding schema.
+def test_node_new_attribute_assignment():
```
WilliamJamieson, line 650: This should be in `test_stnode.py` as it is testing adding to the node object.

WilliamJamieson (changes requested):
A follow on question to the requested change. Can this be done directly on a `DataModel` object itself rather than the `node` object(s) it wraps?

tests/test_models.py
perrygreenfield: So moved.

perrygreenfield: @WilliamJamieson are you asking for it to work only on `DataModel`objects (instead of node objects), or both? If the former, I can imagine many cases where the actions on node objects is desired. So I assume you mean both. I'll take a look at that.

perrygreenfield: In looking at the code, doesn't it already work for DataModel objects? I did a test interactively and it appeared to work fine. Is the issue that a corresponding test is needed?

PR #293, opened 2023-11-16
Move `dqflags` from `romancal` to `roman_datamodels` and make them enums.

WilliamJamieson:
When working on jump detection for `romancal` I noticed that the `dqflags` were stored in `romancal` instead of `roman_datamodels`. The `dqflags` being directly part of the pipeline instead of the datamodels package proved to be an issue for JWST as some tools wished to not have a direct dependency on the JWST pipeline simply to open and interpret existing output files, see spacetelescope/stdatamodels#134. 

This PR begins the process of moving the `dqflags` from `romancal` to `roman_datamodels` by "duplicating" them into `roman_datamodels`. Note that at this time, `romancal` will still be referencing its internally defined flags, an update to this effect can be made later to `romancal` which is seamless to the development process.

Additionally, this PR transitions the `dqflags` from using simple dictionaries to defining them using an `Enum` as I discussed briefly with @schlafly. This adds two useful things:

1. It protects the defined values of the dqflags from accidentally being changed at run time. The current use of a `dict` is susceptible to the possibility of a module importing the dqflags updating a flag's value, which would change that value globally.
2. Allows for `.` access to the flag values e.g. `pixel.GOOD` while still retaining the dictionary like access e.g. `pixel["GOOD"]`. The `.` access is nice because it enables static checkers to "flag" when something like a typo in a flag's name occurs and code completion engine's to suggest the list of possible flags.

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@
 
 - Allow assignment to or creation of node attributes using dot notation of object instances
   with validation. [#284]
+- Move ``dqflags`` from ``romancal`` to ``roman_datamodels``. [#293]
 
 - Bugfix for ``model.meta.filename`` not matching the filename of the file on disk. [#295]
 
--- a/src/roman_datamodels/dqflags.py
+++ b/src/roman_datamodels/dqflags.py
@@ -1,0 +1,68 @@
+""" Roman Data Quality Flags
+
+The definitions are documented in the Roman RTD:
+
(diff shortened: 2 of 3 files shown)


PaulHuwe (approved):
LGTM. Be sure to get approval on the RCAL companion ticket before merging this.

schlafly:
This looks good to me.  It's good to have this in roman_datamodels rather than romancal so that people can use the flags without bringing in romancal as a dependency.

PR #295, opened 2023-11-24
RCAL-723 Fix `model.meta.filename` when saving model

WilliamJamieson:
Resolves [RCAL-723](https://jira.stsci.edu/browse/RCAL-723)

Closes spacetelescope/romancal#1026

When saving a `DataModel` to an arbitrary ASDF file, it was noted that the `model.meta.filename` does not reflect the name of the file it was saved to. This PR makes sure that when saving a `DataModel` to `foo.asdf` that `Datamodel.meta.filename==foo.asdf` in the saved file. 

Note that this solves RCAL-723 when running individual steps or outputting all the step results.

**Checklist**
- [x] updated relevant tests

### DIFF
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -61,6 +61,10 @@ def _temporary_update_filename(datamodel, filename):
 
         yield
         datamodel._instance.meta.filename = old_filename
+        return
+
+    yield
+    return
 
 
 class DataModel(abc.ABC):
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -12,6 +12,7 @@ import copy
 import datetime
 import functools
 import sys
+from contextlib import contextmanager
 from pathlib import Path, PurePath
 
 import asdf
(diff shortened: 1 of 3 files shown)


src/roman_datamodels/datamodels/_core.py
```diff
@@ -48,6 +47,22 @@ def wrapper(self, *args, **kwargs):
...
+    can be saved with that new file name without changing the current model's filename
+    """
+    from roman_datamodels.stnode import Filename
+
+    if "meta" in datamodel._instance and "filename" in datamodel._instance.meta:
+        old_filename = datamodel._instance.meta.filename
+        datamodel._instance.meta.filename = Filename(filename)
+
+        yield
+        datamodel._instance.meta.filename = old_filename
```
WilliamJamieson, line 63: This specifically exists so that the `DataModel` instance being saved via `save` or `to_asdf` does not update its `filename`, the filename is only updated in the newly written ASDF file and the model which can be opened from that.

I chose to do it this way because `AsdfFile.write_to` is employed by these `to_asdf` (which is called by `save`), which writes the `AsdfFile` object to the new file descriptor passed to it, but it does not update the file descriptor of the `AsdfFile` object. This means the asdf file wrapped by the `DataModel` instance is not changing, so the filename of the python object being saved should not be updated.

PaulHuwe:
Could you add a regression test link? I think this may break them (if previous tests were nor careful with file names).

WilliamJamieson:
> Could you add a regression test link? I think this may break them (if previous tests were not careful with file names).

You are quite right, it breaks most of the regression tests. See https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/502/. In reviewing them, it looks like all the failures are due to the `roman.meta.filename` not matching, which points to us not being careful about the names before.

schlafly:
Let's discuss what we want here at a Thursday tag up.  Honestly, it's not always clear to me what one wants for meta.filename.  I guess if you save it you want it to be the filename?  But if you pass it through steps should the steps be updating it to what they would save it to, were they to write it out?  When I was playing with related things last week I got a bit confused about when meta.filename was used to decide what output filename to use, vs. when command line arguments were used, vs. when file names were constructed automatically from the step names.

But let's do this after the upcoming release.

PaulHuwe (approved):
LGTM

nden (approved):

nden:
@WilliamJamieson Feel free to merge once checks pass

PR #301, opened 2024-01-08
Remove duplicate validation, replace assumed validation on `AsdfFile.__init__` with `AsdfFile.validate`

braingram:
This PR removes duplicate validation during `to_asdf`:
https://github.com/spacetelescope/roman_datamodels/blob/70f05e88d2c32a716358d0c2c3fbf8770f600165/src/roman_datamodels/datamodels/_core.py#L219-L223
The assignment to `tree` triggers the first validation and the `write_to` triggers a second.

This PR also updates several uses of `AsdfFile.__init__` to validate tree contents (replacing it with an explicit call to `AsdfFile.validate`).

These changes are to prepare roman_datamodels for an upcoming deprecation in asdf (see https://github.com/asdf-format/asdf/pull/1691).

Regression tests running with no errors: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/515/

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -220,7 +220,7 @@ class DataModel(abc.ABC):
     def to_asdf(self, init, *args, **kwargs):
         with validate.nuke_validation(), _temporary_update_filename(self, Path(init).name):
             asdf_file = self.open_asdf(**kwargs)
-            asdf_file.tree = {"roman": self._instance}
+            asdf_file["roman"] = self._instance
             asdf_file.write_to(init, *args, **kwargs)
 
     def get_primary_array_name(self):
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -39,7 +39,10 @@ def _set_default_asdf(func):
         if self._asdf is None:
             try:
                 with validate.nuke_validation():
-                    self._asdf = asdf.AsdfFile({"roman": self._instance})
+                    af = asdf.AsdfFile()
+                    af["roman"] = self._instance
+                    af.validate()
+                    self._asdf = af
             except ValidationError as err:
                 raise ValueError(f"DataModel needs to have all its data flushed out before calling {func.__name__}") from err
 
@@ -126,8 +129,10 @@ class DataModel(abc.ABC):
                 )
             with validate.nuke_validation():
                 self._instance = init
-                self._asdf = asdf.AsdfFile({"roman": init})
-
+                af = asdf.AsdfFile()
+                af["roman"] = self._instance
+                af.validate()
+                self._asdf = af
                 return
 
         if init is None:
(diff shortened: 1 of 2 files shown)


src/roman_datamodels/datamodels/_core.py
```diff
@@ -38,7 +38,10 @@ def wrapper(self, *args, **kwargs):
         if self._asdf is None:
             try:
                 with validate.nuke_validation():
-                    self._asdf = asdf.AsdfFile({"roman": self._instance})
+                    af = asdf.AsdfFile()
+                    af["roman"] = self._instance
+                    af.validate()
```
WilliamJamieson, line 43: This seems a bit verbose. Is there something wrong with:
```python
self._asdf = asdf.AsdfFile({"roman": self._instance})
self._asdf.validate()
```

braingram: Thanks for taking a look and good question. Unfortunately, the code you proposed would validate the instance twice. Once during `AsdfFile.__init__` and again during `AsdfFile.validate`. Also, the reason for this PR is that asdf is deprecating validation during `AsdfFile.__init__`. The code you proposed would result in a `DeprecationWarning` and test failure when the validation fails. The code proposed in this PR avoids the possibility of the `DeprecationWarning` by not providing the `_instance` during `AsdfFile.__init__.`.

WilliamJamieson: Is asdf deprecating passing a dictionary directly to the `AsdfFile` initializer? If so, then that is a huge change of behavior!

If it's simply raising `DepreciationWarning` because of the validation change then a more thoughtful change should be considered as initializing with a dictionary is an extremely common usage of `AsdfFile`.

I would request that a `validate=True` option should be added to the initializer which preserves the current behavior without deprecation. You can set it so it defaults to `None` and raises the warning of impending behavior changes in that case.

braingram: asdf is not deprecating initializing with a dictionary. However if that dictionary is invalid one will see a `DeprecationWarning` noting that validation on `AsdfFile.__init__` is deprecated and a `ValidationError`. If the dictionary is valid no warning or error will be seen. This was done to allow test suites (like the one in `roman_datamodels` to detect the change (the warning appears during the CI run which gets turned into an error).

PaulHuwe (approved):
LGTM modulo William's comments.

src/roman_datamodels/datamodels/_core.py
braingram: @WilliamJamieson I'm not sure I addressed your comments here. asdf is deprecating (for removal) validation on `AsdfFile.__init__` to reduce the duplication of functionality in asdf. Adding a `validate` keyword that triggers validation overlaps with functionality that is already provided by `AsdfFile.validate`.

WilliamJamieson (approved):

PR #345, opened 2024-04-25
fix mk_outlier_detection

nden:
Resolves [RCAL-823](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes #

This PR addresses an issue with where mk_outlier_detection is placed. Since it's a tagged node it needs to be under tagged_nodes, not common_meta. The first commit repoints roman_datamodels to rad/main and shows the failure. The second fixes it.

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,8 +18,8 @@ dependencies = [
     "gwcs >=0.18.1",
     "numpy >=1.22",
     "astropy >=5.3.0",
-    # "rad @ git+https://github.com/spacetelescope/rad.git",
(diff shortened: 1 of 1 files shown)


nden:
And strangely it fails for me locally but doesn't fail on GA.

PaulHuwe:
> And strangely it fails for me locally but doesn't fail on GA.

The only change here is repointing to dev version of RAD, which had no differences with the released version.

nden:
My bad, I was overwriting rad locally.

PR #346, opened 2024-04-25
add ref_file to catalog metadata

nden:
Resolves [RCAL-823](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes #

This PR addresses ...

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,9 +18,10 @@ dependencies = [
     "gwcs >=0.18.1",
     "numpy >=1.22",
     "astropy >=5.3.0",
-    # "rad @ git+https://github.com/spacetelescope/rad.git",
-    "rad >= 0.19.2",
+    #"rad @ git+https://github.com/spacetelescope/rad.git",
+    #"rad >= 0.19.2",
     "asdf-standard >=1.0.3",
+    "rad @ git+https://github.com/nden/rad.git@d485ddd",
 ]
 dynamic = [
     "version",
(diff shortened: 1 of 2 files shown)


PaulHuwe:
Is the RAD branch mislinked?

schlafly:
Following up from our discussion at tag up, here are two options:
1. We change the 'finalize_result' step of romancal.stpipe to only set ref_files-related things if ref_files exists in the metadata.  I am a little partial to this approach since the L3 / L4 dependence on CRDS is pretty minimal.  It's not as if the aperture corrections are going to be moving around frequently.  But it's not what we decided on the call.
2. We add an additional ref_file_minimal schema that just contains the crds.sw_version and crds.context_used keywords.  We change the L3 and catalog schemas to use this new minimal schema.  This is what we decided on the call.

Thoughts?

nden:
@schlafly I think it's fine to modify `finalize` to only write out the version of CRDS if there were reference files used. This will take care of the L3 products. We should probably still add a section for the L4 products?

schlafly:
I'll defer to you.  I think I would just say that both L3 and L4 files could not have this in the schema and not populate this in romancal.  There is some value to the context and sw_version for L3 and L4 files, but it just seems pretty low to me, and I feel like right now we're primarily motivated by making romancal.stpipe happy rather than actually setting metadata we care about.  But your call!

nden:
Closing as not needed. Will consider again when needed.

PR #349, opened 2024-05-07
Skip QTables in individual image metadata.

schlafly:
This PR adds QTable to the list of things to skip for the individual image metadata.

QTables get made as part of the PSF catalog that is now the tweakreg_catalog, instead of the old ndarray.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@
 
 - Add conversion of dict to string during Qtable construction [#348]
 
+- Do not include QTables in individual image metadata [#349]
+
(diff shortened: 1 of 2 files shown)


schlafly:
@PaulHuwe , I needed this in order to get the IIM to work with the new astropy Table PSF catalogs.  Because those are now an astropy table rather than an ndarray the special casing in IIM was not catching them.  Any objection to merging this?  I think this is the only roman_datamodels update needed.

PaulHuwe (approved):
LGTM. Update the change log, please.

PaulHuwe:
Merging.

PR #352, opened 2024-05-23
RCAL-833 Recursively convert all meta attributes during model casting

stscieisenhamer:
Resolves [RCAL-833](https://jira.stsci.edu/browse/RCAL-833)

This PR addresses issues found while trying to convert TVAC raw/level 1 data into a RampModel.

The issue is that the meta update was using `dict.update`, which doesn't transcribe/copy all the attributes. This left attributes with TVAC tags, producing validation warnings when the RampModel is produced.

Solution is to recursively traverse the meta tree, equating/converting each item.

**Checklist**
- [x] updated relevant tests

### DIFF
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -5,6 +5,7 @@ This module provides all the specific datamodels used by the Roman pipeline.
     the top-level STNode type that the datamodel wraps. This STNode type is derived
     from the schema manifest defined by RAD.
 """
+from collections.abc import Mapping
 
 import asdf
 import numpy as np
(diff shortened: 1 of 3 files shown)


stscieisenhamer:
Attention Reviewers: This is one method; the other is presented in PR #353 

This implementation is fixing the previously implemented way of creating the ramp node, update the node with the values from the input model, then create the RampModel from that node. The previous version of this code suffered conversion issues due to the whole of the `meta` tree was simply using `dict.update`, essentially not touching any of the items.  The solution here does the recursive walk through the attributes, touching all items. Doing so found another special case involving lists/Lnodes. 

The implementation in #353 does everything, but directly on the RampModel and not the Ramp stnode. #353 does solve the issue experienced. However, extra keys, such as `filename` and `model_type` also get modified. To rectify that, the meta update will also need to become recursive to filter out such keys.

schlafly:
@PaulHuwe , can I ask you to take a look at these?  Following a discussion with Jonathan, I think either of these are good.  I have a vague, poorly defined preference for avoiding any code that knows about nodes, driven mostly by my not really understanding what these are good for, which would argue for #353.  On the other hand, I haven't figured out where the existing code avoids updating model_type, so there is some hidden magic via that approach as well---I'm happy either way.  Take a look and then we'll let Jonathan decide?  Thanks!

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -135,37 +137,59 @@ class RampModel(_RomanDataModel):
...
+        Returns
+        -------
+        output_model : RampModel
+            The RampModel built from the input model. If the input is already
+            a RampModel, it is simply returned.
         """
-
         if isinstance(model, cls):
             return model
+        if not isinstance(model, (ScienceRawModel, TvacModel)):
```
PaulHuwe, line 158: Should RampModel be in this list to match the comments?

Also, I think FpsModel should be allowed in this function as well.

tests/test_models.py
```diff
@@ -958,3 +958,10 @@ def test_datamodel_save_filename(tmp_path):
 
     with datamodels.open(filename) as new_ramp:
         assert new_ramp.meta.filename == filename.name
+
+
+@pytest.mark.parametrize("model_class", [datamodels.RampModel, datamodels.ScienceRawModel, datamodels.TvacModel])
+def test_rampmodel_from_science_raw(model_class):
+    """Test creation of RampModel from raw science/tvac"""
+    model = utils.mk_datamodel(model_class)
+    ramp = datamodels.RampModel.from_science_raw(model)
```
PaulHuwe, line 967: Maybe add a line where you set a value and confirm the set value is properly propagated?

src/roman_datamodels/datamodels/_datamodels.py
stscieisenhamer: RampModel is implicitly checked for in the preceding statement. However, no issue with adding here also.

PaulHuwe (approved):
LGTM

ddavis-stsci (approved):
LGTM

PR #353, opened 2024-05-23
 RCAL-833 Recursively convert all meta attributes during model casting (pure model version)

stscieisenhamer:
Resolves [RCAL-833](https://jira.stsci.edu/browse/RCAL-833)

This PR addresses issues found while trying to convert TVAC raw/level 1 data into a RampModel.

The issue is that the meta update was using dict.update, which doesn't transcribe/copy all the attributes. This left attributes with TVAC tags, producing validation warnings when the RampModel is produced.

The solution is to remove the `stnode` direct updating and do the updating through the RampModel directly

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -142,7 +142,7 @@ class RampModel(_RomanDataModel):
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


stscieisenhamer:
Closing this PR in favor of #352

PR #358, opened 2024-07-14
Enable "lazy_tree" for all Datamodels

braingram:
Regtests all pass:
https://github.com/spacetelescope/RegressionTests/actions/runs/10163978364

This PR enables the asdf [lazy_tree](https://asdf.readthedocs.io/en/latest/api/asdf.config.AsdfConfig.html#asdf.config.AsdfConfig.lazy_tree) feature for roman_datamodels. By default asdf set's `lazy_tree=False`, this PR sets the default to `True` for the calls to `roman_datamodels.datamodels.open`.

This PR also changes a few `isinstance` checks to include the "lazy" nodes returned by asdf (for example, instead of a `dict` asdf will return a `AsdfDictNode` when `lazy_tree=True`).

Finally, this PR adds `lazy=True` to the roman datamodels asdf converters to signify that these converters can handle "lazy" objects.

Enabling "lazy_tree" allows asdf to defer conversion of custom objects (like `astropy.unit.Quantity`) until they are accessed (from the containing object). This allows asdf to also defer loading the blocks containing the array data for `Quantity` (this is otherwise impossible due to the handling of the input for `Quantity.__init__`). To provide an example, on roman_datamodels main:

```python
>>> import roman_datamodels as rdm
>>> m = rdm.open("r0000101001001001001_01101_0001_WFI01_cal.asdf")  # one of the regtest files
>>> sum([b.loaded for b in m._asdf._blocks._blocks])
36
```
shows that 36 blocks were loaded from disk when the file was opened (the file contains a total of 41 blocks).
With this PR:
```python
>>> import roman_datamodels as rdm
>>> m = rdm.open("r0000101001001001001_01101_0001_WFI01_cal.asdf")  # one of the regtest files
>>> sum([b.loaded for b in m._asdf._blocks._blocks])
2
```
Only 2 blocks were loaded (asdf does this as a sanity check of the file, loading the first and last block to confirm that the blocks are ordered appropriately).
Furthermore, with this PR, the above example takes 142.8M of RAM, whereas with main the example takes 486.5M.

**Checklist**
- [ ] updated relevant tests

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 
 - replace usages of ``copy_arrays`` with ``memmap`` [#360]
 
+- Enable asdf "lazy_tree" mode for all roman datamodels files [#358]
+
 0.20.0 (2024-05-15)
 ===================
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 dependencies = [
-    "asdf >=3.1.0",
+    "asdf >=3.3.0",
     "asdf-astropy >=0.5.0",
     "gwcs >=0.19.0",
     "numpy >=1.22",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
     "astropy >=5.3.0",
     "rad >= 0.20.0",
     # "rad @ git+https://github.com/spacetelescope/rad.git",
-    "asdf-standard >=1.0.3",
+    "asdf-standard >=1.1.0",
 ]
 dynamic = [
     "version",
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -19,6 +19,7 @@ from pathlib import Path, PurePath
 import asdf
 import numpy as np
 from asdf.exceptions import ValidationError
+from asdf.lazy_nodes import AsdfDictNode, AsdfListNode
 from astropy.time import Time
 
 from roman_datamodels import stnode, validate
@@ -315,10 +316,10 @@ class DataModel(abc.ABC):
         """
 
         def recurse(tree, path=[]):
-            if isinstance(tree, (stnode.DNode, dict)):
+            if isinstance(tree, (stnode.DNode, dict, AsdfDictNode)):
                 for key, val in tree.items():
                     yield from recurse(val, path + [key])
-            elif isinstance(tree, (stnode.LNode, list, tuple)):
+            elif isinstance(tree, (stnode.LNode, list, tuple, AsdfListNode)):
                 for i, val in enumerate(tree):
                     yield from recurse(val, path + [i])
             elif tree is not None:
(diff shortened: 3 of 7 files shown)


PaulHuwe (approved):
LGTM

schlafly (approved):
Wow, I'm surprised that no changes were necessary to romancal proper.

Just out of curiosity, a "block" is an entire data extension, so for files with two blocks, lazy_tree doesn't do very much, even if they're long blocks?  In my imagination a big win for lazy_tree is for operations that only need to read in the metadata (operations like [fits.getheader(x)[field] for x in list_of_filenames]), and I'm trying to understand if that's still the case with files with <= 2 data extensions.

braingram:
> Wow, I'm surprised that no changes were necessary to romancal proper.

There are some follow-up changes that are worth considering related to removing `lazy_load` usage:
https://github.com/search?q=repo%3Aspacetelescope%2Fromancal%20lazy_load&type=code
as that largely defeats the IO benefits from `lazy_tree` (more on that below).

> Just out of curiosity, a "block" is an entire data extension, so for files with two blocks, lazy_tree doesn't do very much, even if they're long blocks? In my imagination a big win for lazy_tree is for operations that only need to read in the metadata (operations like [fits.getheader(x)[field] for x in list_of_filenames]), and I'm trying to understand if that's still the case with files with <= 2 data extensions.

Thanks for the question. I glossed over some details when referring to which blocks were "loaded". There is still a benefit for files with <= 2 data extensions as the above used `block.loaded` property refers to the ASDF block header (not the block data).

To try and add some more (but hopefully not too much) detail. Each ASDF block has a binary [header](https://asdf-standard.readthedocs.io/en/1.1.1/file_layout.html#block-header) and [data](https://asdf-standard.readthedocs.io/en/1.1.1/file_layout.html#block-content). When a file is opened, asdf reads the headers for the first and last blocks but typically does not load the data (when lazy_load=True). However, when the data is accessed (or the converter forces the loading in the case of `astropy.Quantity`) asdf will read and cache the block data (for the block referred to by the array). For this PR since we're enabling lazy_tree (and lazy_load is on by default) opening a roman file with the default arguments will:

```python
>>> m = rdm.open("r0000101001001001001_01101_0001_WFI01_cal.asdf")
>>> sum([b.loaded for b in m._asdf._blocks._blocks])
2
```

Load the first and last block headers. However the data portion of the blocks are not yet read (this uses more private asdf API as the public API for interfacing with the blocks is within the extension code and would complicate the example here).

```python
>>> m._asdf._blocks._blocks[0]._cached_data
None
```

For this particular file, the 0th block corresponds to coefficients of a polynomial transform in the gwcs object. So if we access the wcs it will trigger loading the block data:
```python
>>> m.meta.wcs
>>> m._asdf._blocks._blocks[0]._cached_data
array([...])
```

Finally, to circle back to the relationship with `lazy_load`. If a file is opened with `lazy_load=False` every block header and block data will be loaded when the file is first opened. Using the above example:

```python
>>> m = rdm.open("r0000101001001001001_01101_0001_WFI01_cal.asdf", lazy_load=False)
>>> sum([b.loaded for b in m._asdf._blocks._blocks])
41
>>> m._asdf._blocks._blocks[0]._data  # returns an array for every block in the file
array([...])
```

So for the most part using `lazy_load` cancels any benefits from `lazy_tree`. Let me know if more details would be helpful and if you have any questions.

PR #372, opened 2024-08-05
only use roman.meta for get_crds_parameters

### LINKED ISSUE #366: `DataModel.to_flat_dict` includes arrays when using `include_arrays=False`
If I open a roman file (one of the regression test files for romancal) and use `to_flat_dict(include_arrays=False)` arrays are included in the flat dictionary.
```python
>>> import roman_datamodels as rdm
>>> m = rdm.open("test_bigdata/roman-pipeline/dev/WFI/image/r0000101001001001001_01101_0001_WFI01_cal.asdf")
>>> fd = m.to_flat_dict(include_arrays=False)
>>> fd["roman.meta.dq"]
array([[1048576, 1048576, 1048576, ..., 1048576, 1048576, 1048576],
       [      0,       0,       0, ...,       4,       0,       0],
       [1048576, 1048576, 1048576, ..., 1048576, 1048576, 1048576],
       ...,
       [      0,       0,       0, ...,       0,       0,       0],
       [      0,       0,       0, ...,       0,       0,       0],
       [1048576, 1048576, 1048576, ..., 1048576, 1048576, 1048576]],
      dtype=uint32)
```

### LINKED ISSUE #367: `DataModel.get_crds_parameters` accesses all contents of the file (triggering loading for "lazy" files)
The implementation of `DataModel.get_crds_parameters`:
https://github.com/spacetelescope/roman_datamodels/blob/911485a5913e61a42dc0b79c83325eadec310d3a/src/roman_datamodels/datamodels/_core.py#L338-L342
calls `DataModel.to_flat_dict` which recursively accesses all portions of a file:
https://github.com/spacetelescope/roman_datamodels/blob/911485a5913e61a42dc0b79c83325eadec310d3a/src/roman_datamodels/datamodels/_core.py#L300-L302

This results in loading of all portions of the file when a model is used:
- as an argument to [get_reference_files](https://github.com/spacetelescope/stpipe/blob/ab747f93b0478a0d6400498dd8b9089113e49cb5/src/stpipe/step.py#L822)
- as in input for a step (for [fetching parameters](https://github.com/spacetelescope/stpipe/blob/ab747f93b0478a0d6400498dd8b9089113e49cb5/src/stpipe/step.py#L877))

As `to_flat_dict` is passed `include_arrays=False` (see https://github.com/spacetelescope/roman_datamodels/issues/366 for a related issue) I think it's expected that many portions of the file should not be included (and `DataModel.get_crds_parameters` further filters the results to only a few basic types). Applying this filtering before accessing all aspects of the file (or perhaps only recursing into a portion of the tree and skipping specific attributes like `meta.wcs`) would be a more efficient way to get these parameters.

braingram:
Closes #367
Closes #366 

This PR modifies `DataModel.get_crds_parameters` to only include (and only access) items under `roman.meta`. `get_crds_parameters` is called twice during the setup of commands like the following `strun MyStep mydata.asdf`. Once to get the pars file for the `MyStep` and again to prefetch reference files for `mydata.asdf`. Each of these calls will:
- open the `Datamodel` contains in `mydata.asdf` (each call separately opens the file, the opened datamodel is not shared between calls)
- call `Datamodel.get_crds_parameters`
With the current implementation of `get_crds_parameters`, every attribute of the datamodel is accessed and then the attributes are filtered to include select non-array values (see the code for the specific filtering). For a "lazy" file this will result in loading and converting all contents of the file twice before the file is again re-opened within `MyStep.process`.

With this PR only the nodes under `roman.meta` are accessed and included for `get_crds_parameters`. This PR does not query crds for the required parameter selectors (parkeys) as:
- there doesn't appear to be an API for reliably getting these parkeys
- even with an API some coordination with stpipe would likely be called for as it has a crds dependency (unlike roman_datamodels)

Querying crds would allow fewer metadata entries to be loaded (which is a very minor improvement that might be offset by the crds API) and would protect against the unlikely case that a non-`roman.meta` attribute is used as a selector.

This PR also fixes the "include_arrays" filtering for `DataModel.to_flat_dict`.

Regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/10268370867
show unrelated failures due to background changes.

**Checklist**
- [x] updated relevant tests

### DIFF
--- a/changes/372.bugfix.rst
+++ b/changes/372.bugfix.rst
@@ -1,0 +1,1 @@
+Only use ``roman.meta`` attributes for crds parameter selection.
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -19,7 +19,7 @@ from pathlib import Path, PurePath
 import asdf
 import numpy as np
 from asdf.exceptions import ValidationError
-from asdf.lazy_nodes import AsdfDictNode, AsdfListNode
+from asdf.tags.core.ndarray import NDArrayType
 from astropy.time import Time
 
 from roman_datamodels import stnode, validate
@@ -298,7 +298,9 @@ class DataModel(abc.ABC):
             return val
 
         return {
-            f"roman.{key}": convert_val(val) for (key, val) in self.items() if include_arrays or not isinstance(val, np.ndarray)
+            f"roman.{key}": convert_val(val)
+            for (key, val) in self.items()
+            if include_arrays or not isinstance(val, (np.ndarray, NDArrayType))
         }
 
     def items(self):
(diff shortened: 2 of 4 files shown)


braingram:
I have no opinion about this making it into the current (or next release). If it doesn't make it into this release I'll update the changelog after the release.

schlafly:
Sorry that it's been a while since you filed this.

> With this PR only the nodes under roman.meta are accessed and included for get_crds_parameters. This PR does not query crds for the required parameter selectors (parkeys) as:
> 
> there doesn't appear to be an API for reliably getting these parkeys
> even with an API some coordination with stpipe would likely be called for as it has a crds dependency (unlike roman_datamodels)

Just making sure I follow, you're saying that you could ask CRDS to say what it needs, and then send it back exactly that, rather than just sending the entire meta contents minus the arrays.  I hear you and agree that we shouldn't need to do that and your PR is good as is.

PR #378, opened 2024-08-15
fix mk_level2_image for 3d shape

### LINKED ISSUE #377: shape argument to mk_level2_image doesn't work as described
mk_level2_image(...) has a shape argument that is in part intended to allow level 2 images with different numbers of resultants to be created.  The following lines

https://github.com/spacetelescope/roman_datamodels/blob/5e996c37648bac7fa0ac3c4d5e0ce5b8ddaeb69a/src/roman_datamodels/maker_utils/_datamodels.py#L101-L102

need to be in the opposite order in order for this to work.  Currently mk_level2_image(3, 4088, 4088) makes an amp33 image that is 4088 x 4096 x 128 in size, while 3 x 4096 x 128 is the expected result.  The issue is that n_groups is shape[0], but shape is cut down to shape[1:3] before n_groups is assigned, so n_groups ends up being 4088 in this case at present.

braingram:
I initially hoped to drop the 3d shape input. However the maker_utils tests are quite tied to the current structure and it wasn't immediately obvious how to drop the 3d input without breaking several other tests (which provide a [3d shape](https://github.com/spacetelescope/roman_datamodels/blob/5e996c37648bac7fa0ac3c4d5e0ce5b8ddaeb69a/tests/test_maker_utils.py#L182)).

Instead this PR takes the approach of fixing the 3d shape input handling and adds a regression test for the issue.

Closes #377

Regression tests running at https://github.com/spacetelescope/RegressionTests/actions/runs/10405495032

**Checklist**
- [x] updated relevant tests


schlafly (approved):
LGTM, thanks Brett!

PR #396, opened 2024-10-03
RCAL-853: Give VA scale factor a reasonable default

stscieisenhamer:
WIP [RCAL-853](https://jira.stsci.edu/browse/RCAL-853)

Main PR: [PR#1354](https://github.com/spacetelescope/romancal/pull/1354/files)

This PR gives the `velocity_aberration.scale_facter` meta a default working value of 1.0

## Tasks
- [x] update or add relevant tests
- [x] update relevant docstrings and / or `docs/` page
- [ ] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
  - [ ] [start a `romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/roman_datamodels@<branch>"`)

<details><summary>news fragment change types...</summary>

</details

### DIFF
--- a/src/roman_datamodels/maker_utils/_common_meta.py
+++ b/src/roman_datamodels/maker_utils/_common_meta.py
@@ -256,7 +256,6 @@ def mk_velocity_aberration(**kwargs):
     vab["dec_reference"] = kwargs.get("dec_reference", NONUM)
     vab["scale_factor"] = kwargs.get("scale_factor", 1)
 
-
     return vab
 
 
--- a/src/roman_datamodels/maker_utils/_common_meta.py
+++ b/src/roman_datamodels/maker_utils/_common_meta.py
@@ -254,7 +254,8 @@ def mk_velocity_aberration(**kwargs):
     vab = stnode.VelocityAberration()
     vab["ra_reference"] = kwargs.get("ra_reference", NONUM)
     vab["dec_reference"] = kwargs.get("dec_reference", NONUM)
-    vab["scale_factor"] = kwargs.get("scale_factor", NONUM)
+    vab["scale_factor"] = kwargs.get("scale_factor", 1)
+
 
     return vab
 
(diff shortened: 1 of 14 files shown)


schlafly (commented):
This looks good to me but can you run regtests?  I think those all basically pass through the metadata from the L1s and so probably aren't sensitive to this, but if they are we should know about it.

stscieisenhamer:
The doc issue must be something upstream. I can investigate or leave to others.

src/roman_datamodels/maker_utils/_common_meta.py
```diff
@@ -254,7 +254,7 @@ def mk_velocity_aberration(**kwargs):
     vab = stnode.VelocityAberration()
     vab["ra_reference"] = kwargs.get("ra_reference", NONUM)
     vab["dec_reference"] = kwargs.get("dec_reference", NONUM)
-    vab["scale_factor"] = kwargs.get("scale_factor", NONUM)
+    vab["scale_factor"] = kwargs.get("scale_factor", 1)
```
braingram, line 257: Are these expected to be reasonable? I assumed they were intentionally nonsense values that the pipeline replaces.

PaulHuwe (approved):
LGTM

src/roman_datamodels/maker_utils/_common_meta.py
schlafly: The reason to give a reasonable default here is that we are going to start actually using this value.  The scale factor makes a pretty small effect (fraction of a pixel or so), so you don't really want things to explode if people don't set it.  Something like ra/dec doesn't have a reasonable default that's right to within a fraction of a pixel all of the time, so it's reasonable to put a crazy number there, but lots of people won't want to think about the scale factor and my preference is not to force them to.

schlafly (approved):

src/roman_datamodels/maker_utils/_common_meta.py
braingram: Thanks. It's helpful to know that these are sometimes valid defaults.

PR #417, opened 2024-11-07
Remove assignment validation

braingram:
This PR removes validation on attribute assignment for roman datamodels. This change was discussed during a recent roman calibration tag up and the consensus was that removing this feature will make handling TVAC data easier (and potentially older versions of files once RAD schemas/roman_datamodels are versioned).

With this PR we could consider removing "nuke_validation". The reasoning is that disabling `validate_on_read` now allows reading old/invalid files. Let's say I have a file "foo.asdf" with `FOO_IMAGE` as `exposure_type-1.0.0`. On roman_datamodels main if I try to open this file I get an error:
```python
>> import asdf
>> from roman_datamodels import datamodels as dm
>> m = dm.open("foo.asdf")
ValidationError: 'FOO_IMAGE' is not one of ['WFI_IMAGE', 'WFI_GRISM', 'WFI_PRISM', 'WFI_DARK', 'WFI_FLAT', 'WFI_WFSC']
```
On main disabling `validate_on_read` doesn't prevent this error since roman_datamodels will still validate each assignment. However, with this PR disabling `validate_on_read` allows the file to be opened:
```python
>> cfg = asdf.config.get_config()
>> cfg.validate_on_read = False
>> m = dm.open("foo.asdf")
<roman_datamodels.datamodels._datamodels.ImageModel at 0x161f96890>
```
Since this PR disables assignment validation there's no need to "nuke validation" for assignments. However there would still be a use for this feature if we want to be able to save invalid files.

If a user wants to write an invalid file I think it makes more sense for that file to not use RAD tags/schemas.

Regression tests https://github.com/spacetelescope/RegressionTests/actions/runs/11731951627
show one doctest failure due to the romancal docs testing assignment validation. This failure is addressed in https://github.com/spacetelescope/romancal/pull/1504

## Tasks
- [x] Update or add relevant `roman_datamodels` tests.
- [x] Update relevant docstrings and / or `docs/` page.
- [x] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [x] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [x] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

</details

### DIFF
--- a/src/roman_datamodels/validate.py
+++ b/src/roman_datamodels/validate.py
@@ -7,15 +7,10 @@ import warnings
 from contextlib import contextmanager
 from textwrap import dedent
 
-from asdf import AsdfFile
 from asdf import schema as asdf_schema
-from asdf import yamlutil
-from asdf.exceptions import ValidationError
-from asdf.util import HashableDict
 
 __all__ = [
     "ValidationWarning",
-    "value_change",
 ]
 
 ROMAN_VALIDATE = "ROMAN_VALIDATE"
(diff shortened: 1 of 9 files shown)


src/roman_datamodels/validate.py
```diff
@@ -1,191 +0,0 @@
...
-    error = str(error)
-    if len(error) > 2000:
-        error = error[0:1996] + " ..."
-    errfmt = "While validating {} the following error occurred:\n{}"
-    errmsg = errfmt.format(name, error)
-    return errmsg
-
-
-@contextmanager
-def nuke_validation():
```
WilliamJamieson, line 160: I'm not sure we want to remove this feature yet. It was made so that IPAC and other external devs could have a code independent method (environment variable) to turn off validation.

Yes, this PR removes the validation on assignment but `nuke_validation` also turns off validation on both file open and file write.

braingram: Thanks for taking a look. With this PR disabling [validation_on_read](https://asdf.readthedocs.io/en/latest/asdf/config.html#validate-on-read) allows opening invalid files.

This leaves only writing invalid files (which is possible with `nuke_validation`). Would you provide some more details about when saving a knowingly invalid file is useful? I'd like to better understand why this is useful.

WilliamJamieson: Unfortunately, the `validation_on_read` option does satisfy the request that `nuke_validation` was created for. In that, they wanted a method to disable validation on a code base without having to make changes to that code base itself. `validation_on_read` requires one to actively change an argument within the code to ASDF in order to use while `nuke_validation` allows for one to disable setting an environment variable on their system.

As to why they wanted the ability to write invalid files? That is because they were starting with invalid files and simply wanted to save the results of their code into a new version of the input file. They did not want to deal with fixing whatever was wrong to cause the input file to be invalid just to save what they cared about.

The original motivation for the request was constant expressed frustration with "ASDF" always throwing "useless" validation errors. The real issue was that RAD is/was under fairly constant flux meaning that their input files were constantly becoming invalid for for reasons that were irrelevant to what they were doing and the constant updating and distribution of new "valid" input files was causing a lot of issues. This maybe less of an issue now that RAD is more "stable", but I think it should remain until we "release" a final version of all the RAD schemas (any changes to RAD schemas result in a new schema with a version bump to the given schema).

schlafly:
I'm happy when you and William are.  There's some magic in the current code where knowledge about how using dot notation triggers validation and dictionary notation does not, and a nice feature of this is that that magic will go away, I think?

mairanteodoro (commented):
Looks good! Thanks, @braingram!
Just one question: will the validation still happen on saving the file?

braingram:
> Looks good! Thanks, @braingram! Just one question: will the validation still happen on saving the file?

Thanks! Yeah, asdf will always validate before saving (although roman_datamodels prevents this when "nuke validation is enabled allowing saving invalid files).

PaulHuwe (approved):
LGTM. Can you wait to merge until @WilliamJamieson approves?

WilliamJamieson (approved):
These changes are fine with me. More cleanup can probably be done but that is outside the scope of this PR.

PR #429, opened 2024-11-22
Some updates to the tox config

WilliamJamieson:
This PR updates the tox config slightly so that it works with `tox-uv`.

## Tasks

</details


.github/workflows/ci.yml
```diff
@@ -37,5 +37,5 @@ jobs:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@924441154cf3053034c6513d5e06c69d262fb9a6  # v1.13.0
     with:
       envs: |
-        - linux: withromancal
```
braingram, line 40: Would you restore this? It's required to test opening an association as a `ModelLibrary`:
https://github.com/spacetelescope/roman_datamodels/blob/ab68c3537278140607b7c290fb5fc3cf81a45cd9/tests/test_open.py#L245-L246

WilliamJamieson: I just changed the name

braingram: The `toxini` also now no longer includes `withromancal` (which did not run romancal tests). Here's the run on main:
https://github.com/spacetelescope/roman_datamodels/actions/runs/11978300995/job/33398032820#step:10:398

Where is `open_asn` tested now that there is no `withromancal` tox environment?

tox.ini
```diff
@@ -38,21 +40,24 @@ pass_env =
...
 commands_pre =
-    oldestdeps: minimum_dependencies roman_datamodels --filename requirements-min.txt
-    oldestdeps: pip install -r requirements-min.txt
-    devdeps: pip install -r requirements-dev.txt -U --upgrade-strategy eager
-    pip freeze
+    {list_dependencies_command}
 commands =
     pytest \
-    warnings: -W error \
+    romancal: -W ignore::DeprecationWarning -W ignore::UserWarning -W ignore::SyntaxWarning -W ignore::pytest.PytestUnknownMarkWarning -W ignore::pytest.PytestUnraisableExceptionWarning -W ignore::erfa.core.ErfaWarning -W ignore::roman_datamodels.datamodels._utils.FilenameMismatchWarning -W ignore::astropy.io.fits.verify.VerifyWarning -W ignore::RuntimeWarning --pyargs romancal \
```
braingram, line 60: This downstream test is picking up the pytest configuration for roman_datamodels instead of the pytest configuration for romancal:
```
2024-11-22T21:24:48.5571495Z [1m============================= test session starts ==============================[0m
2024-11-22T21:24:48.5573597Z platform linux -- Python 3.12.7, pytest-8.3.3, pluggy-1.5.0
2024-11-22T21:24:48.5574444Z cachedir: .tox/romancal/.pytest_cache
2024-11-22T21:24:48.5575280Z rootdir: /home/runner/work/roman_datamodels/roman_datamodels
2024-11-22T21:24:48.5576163Z configfile: pyproject.toml
```
I'm guessing this is the need for this long list of warning filters. Would you update the configuration to use a clone of romancal instead of `pyargs`? Similar to the suggestion in https://github.com/spacetelescope/gwcs/pull/530#issuecomment-2488827411

.github/workflows/ci.yml
WilliamJamieson: Oh I see, why don't you just install romancal with the normal tox test.

tox.ini
WilliamJamieson: Its because I am using tox-uv.

.github/workflows/ci.yml
braingram: Thanks for the suggestion. I was concerned that would mask potential issues where roman_datamodels could silently depend on something in romancal. Since romancal isn't a dependency most of the tests should pass without romancal (and testing under those conditions validates that). It's only this test that we need romancal for (at the moment) so running it in a separate environment made sense. I'm open to suggestions on other ways to organize this and all in favor of adding a downstream romancal test but just want to make sure we're also testing `open_asn`.

tox.ini
braingram: Could you elaborate? Does `tox-uv` handle warnings differently?

.github/workflows/ci.yml
WilliamJamieson: Fine, I changed it to an optional

tox.ini
WilliamJamieson: This was not working for me and I didn't feel like fixing it.
https://github.com/spacetelescope/roman_datamodels/blob/ab68c3537278140607b7c290fb5fc3cf81a45cd9/tox.ini#L81

WilliamJamieson: This also uncovered the issue I fixed here: spacetelescope/romancal#1536

braingram: Would you add some description of what the issue was? That would be helpful for whoever reviews/sees that PR.

I am concerned about maintaining this long list of warnings that are dependent on romancal. If a romancal PR adds a change that introduces a new warning (which isn't currently checked for romancal) than this job in roman_datamodels will start failing until we update this list.

If the above command (which solves the maintanence issue by using the filters of the downstream package) wasn't compatible with `tox-uv` why are we switching to it?

WilliamJamieson: Do not worry about it. I took care of it in my last push.

tox.ini
```diff
@@ -32,27 +36,37 @@ description =
...
+    devdeps: -r requirements-dev.txt
+change_dir =
+    romancal: {env_tmp_dir}
 commands_pre =
-    oldestdeps: minimum_dependencies roman_datamodels --filename requirements-min.txt
-    oldestdeps: pip install -r requirements-min.txt
-    devdeps: pip install -r requirements-dev.txt -U --upgrade-strategy eager
-    pip freeze
+    {list_dependencies_command}
+    git clone {[main]rcal_repo}
```
WilliamJamieson, line 65: Should be able to use `git archive --remote={[main]rcal_repo} main pyproject.toml | tar -d` but I couldn't get `git archive` to connect for some reason.

WilliamJamieson: GitHub apparently does not support archive over https.

Used a `git clone -n --depth=1 --filter=blob:none {[main]rcal_repo}` instead this should only get the `.git` without any blob files for the last commit. I can then checkout the single file I care about.

braingram:
What are the benefits of `tox-uv`? Given that uv isn't at 1.0.0 I'm hesitant to start using it in the CI.

.github/workflows/ci.yml
```diff
@@ -25,17 +25,36 @@ jobs:
...
+
+  environment:
+    runs-on: ubuntu-latest
+    steps:
+      - id: data_path
+        run: echo "path=${{ runner.temp }}/data" >> $GITHUB_OUTPUT
+    outputs:
+      data_path: ${{ steps.data_path.outputs.path }}
+  crds_contexts:
+    uses: ./.github/workflows/contexts.yml
```
zacharyburnett, line 46: we should use the one at CRDS instead (to outsource the maintenance there)
```suggestion
    uses: spacetelescope/crds/.github/workflows/contexts.yml@master
```

tox.ini
```diff
@@ -32,27 +36,37 @@ description =
...
     CI
     TOXENV
     CODECOV_*
+set_env =
+    devdeps: UV_INDEX = https://pypi.anaconda.org/liberfa/simple https://pypi.anaconda.org/astropy/simple https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
+    devdeps: UV_INDEX_STRATEGY = unsafe-any-match
 extras =
     test
+uv_resolution =
+    oldestdeps: lowest-direct
```
braingram, line 53: This isn't equivalent to `minimum_dependencies`. Testing locally I can reduce the lower numpy to `0.01` and the environment doesn't fail due to `uv` resolving this to a later version of numpy (1.22) and not throwing an error. Would you switch this back to `minimum_dependencies`.

braingram (changes requested):
Thanks for working on this. One requested change related to the minimum dependencies testing.

tox.ini
WilliamJamieson: No I plan on ending support for minimum_dependencies in the future.

braingram: I see no mention of any end of life for https://github.com/OpenAstronomy/minimum_dependencies and it's in use by other packages (like [sunpy](https://github.com/sunpy/sunpy/blob/644308816c9ad8443cdf8cc94a9e7fd428286e2e/tox.ini#L71)) that are interested in keeping it alive. Given that it's a better test of our lower pins I say we keep using it.

zacharyburnett: I can do 
```console
uv pip install . --resolution lowest-direct
```
and it attempts to install `numpy==1.22.0`:
```
Resolved 20 packages in 116ms
  × Failed to download and build `numpy==1.22.0`
  ├─▶ Build backend failed to determine requirements with `build_wheel()` (exit
  │   status: 1)

  │   [stderr]
  │   Traceback (most recent call last):
  │     File "<string>", line 8, in <module>
  │       from setuptools.build_meta import __legacy__ as backend
  │     File
  │   "/Users/zburnett/.cache/uv/builds-v0/.tmpy0oCfs/lib/python3.13/site-packages/setuptools/__init__.py",
  │   line 10, in <module>
  │       import distutils.core
  │   ModuleNotFoundError: No module named 'distutils'

  ╰─▶ distutils was removed from the standard library in Python 3.12. Consider
      adding a constraint (like `numpy >1.22.0`) to avoid building a version of
      numpy that depends on distutils.
  help: `numpy` (v1.22.0) was included because `roman-datamodels`
        (v0.22.1.dev4+gab68c35) depends on `numpy`
```

braingram: Thanks @zacharyburnett is there a way to translate that failure to the CI so that the `oldestdeps` job will fail if our minimum pins are not correct?

With this PR if I set the lower numpy pin to 0.01:
https://github.com/spacetelescope/roman_datamodels/pull/430/commits/22e67d66d0d8bb04d06fc5a6f43cd72998e7ba66
the oldestdeps job installs `1.22`:
https://github.com/spacetelescope/roman_datamodels/actions/runs/12013414624/job/33486895903?pr=430#step:10:46
and passes making that job much less useful.

zacharyburnett: I believe that might be because Python 3.10 only supports `numpy>=1.22.0`, right? So the lowest possible version of Numpy is `1.22.0`, even if you specify `0.01`

WilliamJamieson: I don't understand what you mean by

> our minimum pins are not correct

The `numpy` dependency is in the `pyporoject.toml` because `roman_datamodels` directly imports it. The minimum pin `roman_datamodels` lists really only matters for its own uses of `numpy`. If some upstream dependency raises its lower pin that does not really matter to `roman_datamodels` as any install should be the transitive closure the dependencies, which is what `uv` is doing, only it targets the lowest version rather than maximum version of each dependency in this case.

Note, I did not write `minimum_dependencies` so that it would exactly "check minimum" pins in the way you are implying, namely that they are exactly correct. I wrote it so that one could check a minimum dependency version install passed the tests. This was so that we could identify via test failures the minimum versions of dependencies within the package being tested.

What you are referencing seems to indicate that you want to test that the minimum dependency listed within a given package is `>=` the transitive closure of the minimum pins in all the other dependencies. I don't think this is particularly useful considering any install should be done via a valid transitive closure rather than some exact pinning scheme. All the exact pinning scheme does is create work for your package due to pinning changes upstream. The closure should take care of finding the minimum valid version, the package's pin itself should only serve to indicate the minimum version its own internal code enforces.

Note `minimum_dependencies` only relies on the exact pinning scheme because I did want to implement and support the algorithms necessary to determine the transitive closure of the minimum versions for all the dependencies in general.

tox.ini
```diff
@@ -32,27 +36,40 @@ description =
...
+    devdeps: -r requirements-dev.txt
+    oldestdeps: minimum_dependencies
+change_dir =
+    romancal,oldestdeps: {env_tmp_dir}
 commands_pre =
     oldestdeps: minimum_dependencies roman_datamodels --filename requirements-min.txt
-    oldestdeps: pip install -r requirements-min.txt
-    devdeps: pip install -r requirements-dev.txt -U --upgrade-strategy eager
-    pip freeze
+    oldestdeps: uv pip install -r requirements-min.txt
```
WilliamJamieson, line 66: @braingram This forces an exact pin to all the lower pins. Does this do what you wanted?

.github/workflows/ci.yml
```diff
@@ -25,17 +25,36 @@ jobs:
...
+        CRDS_PATH: ${{ needs.environment.outputs.data_path }}/crds
+        CRDS_SERVER_URL: https://roman-crds.stsci.edu
+        CRDS_CLIENT_RETRY_COUNT: 3
+        CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
+      cache-path: ${{ needs.environment.outputs.data_path }}/crds
+      cache-key: crds-${{ needs.crds_contexts.outputs.roman }}
       envs: |
-        - linux: withromancal
+        - linux: romancal
           coverage: codecov
```
braingram, line 60:
```suggestion
```
Since this job is now running the romancal tests I say we don't include those in the coverage measurement.

tox.ini
braingram: Thanks. Testing locally the `oldestdeps` now fails if one of the lower pins is unresolvable.

braingram (approved):
LGTM. Thanks!

PR #436, opened 2024-11-27
Add default constructor to datamodels and deprecate the maker_utils

WilliamJamieson:
This PR adds the `type(datamodel).create_default(**kwargs)` instance constructor to each of the datamodel types. This is equivalent to calling  `maker_utils.mk_datamodel(type(datamodel), **kwargs)`. This prevents users from accidentally working directly with un-wrapped node objects, external users should be working through the datamodel objects not the nodes. This also simplifies the method to generate a "default" model.

This is the first step in moving the handling of datamodel creation in `romancal`. The next step will be adding constructor methods construct datamodels from datamodels of a different type where this makes sense.

## Tasks

</details


mairanteodoro (commented):
Looks good to me!

PR #440, opened 2024-12-05
RCAL-861: by default compress all arrays

braingram:
Closes https://jira.stsci.edu/browse/RCAL-861

The above ticket description describes setting compression per array. However the discussion in the comments mentions that a "compress everything" approach is good enough for now and we can revisit per-array compression later (which would involve addressing some structural limitations in roman_datamodels).

This PR updates roman_datamodels to compress all array data when writing a file using `lz4` compression. The default compression can be overridden on save (assuming `model` is a `DataModel`):
```python
>>> model.save("foo.asdf", all_array_compression=None)  # for no compression
>>> model.save("foo.asdf", all_array_compression="bzp2")  # for bzip
```
As this builds off the compression infrastructure provided by asdf any algorithms supported by asdf (or asdf extensions) are supported:
https://asdf.readthedocs.io/en/latest/asdf/arrays.html#compression

Regtests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/12317452058

Regtests that intentionally break all compare_asdf calls (to trigger output files to be uploaded to artifactory) at: https://github.com/spacetelescope/RegressionTests/actions/runs/12188485915
and output files at:
https://bytesalad.stsci.edu/ui/repos/tree/General/roman-pipeline-results/regression-tests/runs/2024-12-11_GITHUB_CI_Linux-X64-py3.11-1447

Summary table with a few select files (using the 'truth' file as the uncompressed file).

| filename | uncompressed (MB) | compressed (MB) | compressed percent |
| - | - | - | - |
| r0099101001001001001_r274dp63x31y81_prompt_F158_coadd.asdf | 339.32 | 267.19 | 78.74 |
| r0000101001001001001_0001_wfi01_cal.asdf | 459.82 | 385.93 | 83.93 |
| r0000101001001001001_0001_wfi01_photom.asdf | 459.77 | 385.89 | 83.93 |
| r0000101001001001001_0003_wfi01_rampfit.asdf | 455.76 | 339.83 | 74.56 |
| r0000101001001001001_0001_wfi01_shift_tweakregstep.asdf | 459.79 | 385.90 | 83.93 |
| r0099101001001001001_F158_visit_coadd.asdf | 485.10 | 416.08 | 85.77 |
| r0000101001001001001_0001_wfi01_refpix.asdf | 1535.51 | 248.38 | 16.18 |
| r0000101001001001001_0001_wfi01_linearity.asdf | 1535.51 | 473.78 | 30.85 |
| r0000101001001001001_0001_wfi01_saturation.asdf | 1516.76 | 248.18 | 16.36 |
| r0000101001001001001_0001_wfi01_rampfit.asdf | 460.76 | 339.28 | 73.63 |
| r0000101001001001001_0001_wfi01_dqinit.asdf | 1516.76 | 248.05 | 16.35 |
| r0000101001001001001_0001_wfi01_flat.asdf | 459.77 | 385.89 | 83.93 |
| r0000101001001001001_0001_wfi01_darkcurrent.asdf | 1535.51 | 646.69 | 42.12 |
| r0000101001001001001_0001_wfi01_cal_repoint.asdf | 459.78 | 385.89 | 83.93 |
| r0000201001001001001_0001_wfi01_cal_repoint.asdf | 460.77 | 339.35 | 73.65 |
| r0000201001001001001_0003_wfi01_rampfit.asdf | 455.76 | 339.88 | 74.57 |
| r0000201001001001001_0001_wfi01_cal.asdf | 460.79 | 339.36 | 73.65 |
| r0000201001001001001_0001_wfi01_dqinit.asdf | 1516.76 | 248.04 | 16.35 |
| r0000201001001001001_0001_wfi01_saturation.asdf | 1516.76 | 248.17 | 16.36 |
| r0000201001001001001_0001_wfi01_rampfit.asdf | 460.76 | 339.34 | 73.65 |
| r0000201001001001001_0001_wfi01_flat.asdf | 460.77 | 339.35 | 73.65 |
| r0000101001001001001_0001_wfi01_ALL_SATURATED_cal.asdf | 332.77 | 1.34 | 0.40 |
| r0000101001001001001_0001_wfi01_star.asdf | 459.81 | 385.92 | 83.93 |

## Tasks

</details

### DIFF
--- a/changes/440.feature.rst
+++ b/changes/440.feature.rst
@@ -1,0 +1,1 @@
+Change default compression to lz4.
--- a/changes/440.feature.rst
+++ b/changes/440.feature.rst
@@ -1,1 +1,1 @@
-Change default compression to lz4.
+Change default compression to lz4. The previous default was no compression.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ classifiers = [
 ]
 dependencies = [
     "asdf >=3.3.0",
-    "lz4",
+    "lz4 >= 4.3.0",
     "asdf-astropy >=0.5.0",
     "gwcs >=0.19.0",
     "numpy >=1.24",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ classifiers = [
 ]
 dependencies = [
     "asdf >=3.3.0",
+    "lz4",
     "asdf-astropy >=0.5.0",
     "gwcs >=0.19.0",
     "numpy >=1.24",
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -233,6 +233,8 @@ class DataModel(abc.ABC):
         with validate.nuke_validation(), _temporary_update_filename(self, Path(init).name):
             asdf_file = self.open_asdf(**kwargs)
             asdf_file["roman"] = self._instance
+            if "all_array_compression" not in kwargs:
+                kwargs["all_array_compression"] = "zlib"
             asdf_file.write_to(init, *args, **kwargs)
 
     def get_primary_array_name(self):
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -230,12 +230,11 @@ class DataModel(abc.ABC):
             return asdf.AsdfFile(init, **kwargs)
 
     def to_asdf(self, init, *args, **kwargs):
+        all_array_compression = kwargs.pop("all_array_compression", "lz4")
         with validate.nuke_validation(), _temporary_update_filename(self, Path(init).name):
             asdf_file = self.open_asdf(**kwargs)
             asdf_file["roman"] = self._instance
-            if "all_array_compression" not in kwargs:
-                kwargs["all_array_compression"] = "lz4"
-            asdf_file.write_to(init, *args, **kwargs)
+            asdf_file.write_to(init, *args, all_array_compression=all_array_compression, **kwargs)
 
     def get_primary_array_name(self):
         """
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -234,7 +234,7 @@ class DataModel(abc.ABC):
             asdf_file = self.open_asdf(**kwargs)
             asdf_file["roman"] = self._instance
             if "all_array_compression" not in kwargs:
-                kwargs["all_array_compression"] = "zlib"
+                kwargs["all_array_compression"] = "lz4"
             asdf_file.write_to(init, *args, **kwargs)
 
     def get_primary_array_name(self):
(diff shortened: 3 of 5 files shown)


changes/440.feature.rst
```diff
@@ -0,0 +1 @@
+Change default compression to lz4.
```
PaulHuwe, line 1: Maybe expand this to indicate that compression by default starts here?

PaulHuwe (approved):
LGTM, minus one request. I would like @schlafly 's approval before mergin.

changes/440.feature.rst
braingram: Thanks! How does the updated changelog in [1ff39c2](https://github.com/spacetelescope/roman_datamodels/pull/440/commits/1ff39c2616339adb8d98130e85bb25ba56ba4ae9) look?
```
Change default compression to lz4. The previous default was no compression.
```

PaulHuwe: Looks good.

schlafly (approved):
This looks great!

Maybe some comments about the compression ratios:
Looking at the uncompressed data, the 'ramp' products are 1.5 GB and the 'image' products are 0.5 GB (roughly).  After compression, instead the 'ramp' products are 0.25 - 0.5 GB and the 'image' products are 0.3 - 0.4 GB.

Maybe this makes some sense?  The data are all ultimately generated from some underlying L1 files that are more like ~0.3 GB than 1.5 GB, so it makes sense that compression can take us back to closer to the original size.  The image products represent the ramp products in a lossy way, so in an information theoretic sense we'd expect them to be smaller than the ramp products, but they also add some error and variance arrays that are compressed individually, so it makes sense that they inflate the array sizes as well.

Maybe one thing that would be nice to know: it's our expectation that e.g. the DQ image compresses very well, so that if we added more extensions of that kind it wouldn't "cost" very much.  Is that accurate?

braingram:
> Maybe one thing that would be nice to know: it's our expectation that e.g. the DQ image compresses very well, so that if we added more extensions of that kind it wouldn't "cost" very much. Is that accurate?

Thanks! Good question. To pick on example `r0000101001001001001_0001_wfi01_cal.asdf` it contains a number of arrays (not all shown below assuming var_rnoise is like var_poisson etc). The ASDF blocks store the `allocated_size` (size on disk) and `data_size` (uncompressed size) in the header. If we take the compression ratio as `data_size / allocated_size` than the arrays have the following ratios:

| name | ratio | shape | dtype |
| - | - | - | - |
| data | 0.99 | 4088 x 4088 | float32 |
| dq | 136 | 4088 x 4088 | unit32 |
| err | 0.99 | 4088 x 4088 | float32 |
| var_poisson | 0.99 | 4088 x 4088 | float32 |
| amp33 | 254 | 10 x 4096 x 128 | uint16 |
| border_ref_pix_left | 0.99 | 10 x 4096 x 4 | float32 |
| dq_border_ref_pix_left | 235 | 4096 x 4 | uint32 |

So for dq uncompressed it's ~63 MB, compressed it's ~0.5 MB.

It's safe to say all the disk saving we're getting is from the fixed point arrays (for these files). I expect this pattern (no significant compression ratio) would hold for any floating array unless it has large uniform regions (like could be the case for resampled mosaics which might contain swaths of nans).

PR #455, opened 2025-01-30
RCAL-965 Provide conversion from TVAC/FPS models to ScienceRawModel

stscieisenhamer:
Resolves [RCAL-965](https://jira.stsci.edu/browse/RCAL-965)

This PR addresses the issue that TVAC/FPS data cannot be run through the pipeline, mostly due to the fact that the TVAC-related datamodels are frozen. A new method, `ScienceRawModel.from_tvac_raw` is introduced to do the conversion.

## Tasks

</details


src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -130,6 +130,58 @@ class ImageModel(_RomanDataModel):
...
+        if not isinstance(model, ALLOWED_MODELS):
+            raise ValueError(f"Input must be one of {ALLOWED_MODELS}")
+
+        # Create base ramp node with dummy values (for validation)
+        from roman_datamodels.maker_utils import mk_level1_science_raw
+
+        raw = mk_level1_science_raw(shape=model.shape)
+
+        # Define how to recursively copy all attributes.
+        def node_update(raw, other):
```
braingram, line 161: Does this address the schema differences?

For example, romancal expects `meta.guide_star.window_xstart`:
https://github.com/spacetelescope/romancal/blob/8422362dc33e0d4b3dc6702274b07c740acda37c/romancal/dq_init/dq_init_step.py#L55 (it's probably a bug that it gets this from the input and not the RampModel but either way).

This is ok for the current guidestar schema:
https://github.com/spacetelescope/rad/blob/50958b8991c38ab0071b8a1e9b2297e329444a1b/src/rad/resources/schemas/guidestar-1.0.0.yaml#L76
but the TVAC guidestar schema doesn't contain that data (it appears to be `gw_window_xstart`:
https://github.com/spacetelescope/rad/blob/50958b8991c38ab0071b8a1e9b2297e329444a1b/src/rad/resources/schemas/tvac/guidestar-1.0.0.yaml#L57

stscieisenhamer: This is a good point! Guiding stuff was not such a concern, but will go through and take a closer look. Got any good schema differs?

braingram: I haven't used any schema differs before but would be curious if you find one you like.

schlafly:
Just trying to figure out how this works... if I do strun roman_elp tvac.asdf, when or where does from_tvac_raw get called?

Re schema differences, I think we should try to hit the things that romancal looks for but we don't need to copy everything over just because.  The guide window xstart is an interesting case.  romancal looks for it, but TVAC doesn't actually use guide windows and so if the code runs I would be fine with it without messing with guide window things.

src/roman_datamodels/datamodels/_datamodels.py
stscieisenhamer: Well, no differs found. However, after discussion with Eddie, primary goal is to cover only absolutely necessary meta so that ELP will run. I believe those are covered. If not, they will be handled in other issues.

stscieisenhamer:
@schlafly This is all invoked in the dq_init_step. See [romancal PR#1596](https://github.com/spacetelescope/romancal/pull/1596)

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -130,6 +130,58 @@ class ImageModel(_RomanDataModel):
...
+            """Implement update to directly access each value"""
+            for key in other.keys():
```
PaulHuwe, line 163: This is rather similar to:
https://github.com/spacetelescope/roman_datamodels/blob/main/src/roman_datamodels/datamodels/_datamodels.py#L177

and I seem to recall other spots in the code doing similar recursive updating. Maybe it is best to make a utility function that these methods can call?

PaulHuwe (changes requested):
I am a little surprised more changes aren't needed. I notice that there are regression test failures - are these expected?

schlafly:
Thanks, I think this looks reasonable.  Can you add some documentation of what exactly we're trying to accomplish here, though?  Probably as a docstring in from_tvac_raw, but potentially also in the RTD if there's a reasonable place.

i.e., something like:
romancal supports processing a selection of files which use an outdated schema.  It supports these with a bespoke method that converts the files to the new format when they are read in dq_init.  This conversion does not do a detailed mapping between all of the new and old metadata, but instead opportunistically looks for fields with common names and assigns them.  Other metadata with non-matching names is simply copied in place.  This allows processing to proceed and preserves the original metadata, but the resulting files have duplicates of many entries.
If that sounds right?

I think the biggest practical issue is that a lot of the guide window stuff changed name and we'll now have those bogus fields twice, which is ugly but practical.  That said, since guide windows aren't actually being used for TVAC---I think?  @tddesjardins ? --- it's hard for me to justify writing code to do the conversion of all of the unused fields.

Have you reviewed the schema to see if that's the biggest area where things got renamed?  I went through Tyler's L2 schema doc to try to make sure, and I think that's the case, but it would be good to have another set of eyes on it.

src/roman_datamodels/datamodels/_datamodels.py
stscieisenhamer: And methinks have accomplished it.

stscieisenhamer:
@schlafly I've updated the docstrings in both the roman_datamodel's api and the romancal high-level description of the dq_init step.

stscieisenhamer:
For all reviewers: I believe all the comments have been addressed. Ready for full review

src/roman_datamodels/datamodels/_utils.py
```diff
@@ -23,6 +24,78 @@ class FilenameMismatchWarning(UserWarning):
     """
 
 
+def _node_update(to_node, from_node, extras=None, extras_key=None, ignore=None):
+    """Copy node contents from an existing node to an existing node
```
PaulHuwe, line 28:
```suggestion
    """Copy node contents from an existing node to another existing node
```

src/roman_datamodels/datamodels/_utils.py
```diff
@@ -23,6 +24,78 @@ class FilenameMismatchWarning(UserWarning):
...
+
+    How the copy occurs depends on existence of keys in `to_node`
+
+    If key exists in `to_node`, contents is converted from `from_node` stnode type to
```
PaulHuwe, line 32:
```suggestion
    If key exists in `to_node`, contents are converted from `from_node` stnode type to
```

PaulHuwe (changes requested):
LGTM outside of the one question I posed

src/roman_datamodels/datamodels/_utils.py
```diff
@@ -23,6 +24,78 @@ class FilenameMismatchWarning(UserWarning):
...
+
+    for key in from_node.keys():
+        if key in ignore:
+            continue
+        if key in extras:
+            new_extras[key] = from_node[key]
+            continue
+        if key in to_node:
+            if isinstance(to_node[key], Mapping):
+                _node_update(getattr(to_node, key), getattr(from_node, key), extras=extras, extras_key=extras_key)
```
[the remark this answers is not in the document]
stscieisenhamer: True, and this can happen with the `ignore` option also. Will work that. Methinks that the parameters will become fully specified attributes, such as `meta.statistics`. Further thoughts welcomed.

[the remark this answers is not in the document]
stscieisenhamer: Updated so that the heirarchy is preserved. Everything now goes into a structure like `model.extras...`. So, for the statistics, it is `models.extras.tvac.meta.statistics`.

Thoughts?

PaulHuwe (approved):
Approved modulo test failures (likely just requiring a rebase).
I like the new function!

PR #458, opened 2025-02-05
Switch to Using Explicit Lazy nodes

WilliamJamieson:
This PR accomplishes several  goals:

1. Replacing the `maker_utils` with something more useful.
2. Preventing the unnecessary creation of filler data until it is absolutely necessary. This should reduce the footprint of DataModels during the creation and manipulation up until they are written, meaning that it should be less costly to create a convenience data model if one requires it.
3. Simplifies the data model API. One just creates a data model, there is no extra steps like "use this external constructor" or "use this special method".
4. Making the data models / nodes explicit in there Python code rather than dynamic. This should aid with static analysis tools and ides.

We currently have the `maker_utils`, which effectively lay out the structure of an object using a bunch of nested dictionaries. Consider the util for making a `WfiImage` node:
https://github.com/spacetelescope/roman_datamodels/blob/6fb23b57916f801f1614af5b7c751c8769d704ed/src/roman_datamodels/maker_utils/_datamodels.py#L68-L142

which is basically laying a "`WfiImage`" object but using a dictionary and then calling other functions like
https://github.com/spacetelescope/roman_datamodels/blob/6fb23b57916f801f1614af5b7c751c8769d704ed/src/roman_datamodels/maker_utils/_common_meta.py#L428-L447

to nest further dictionary like things that are defining an object. This means that for us to create data models that we can then write, without validation issues, to an ASDF file, we are coding something like a Python object (through a dictionary) that describes the data detailed in the schema.

The theory of this PR is that instead of coding it using a dictionary and a dynamically generated type, we should just code a python object representing the data object described by ``rad``. The code and mechanisms here are to simplify what needs to be written into these objects as much as possible while maintaining our abilities to call writable data models into existence. In essence, the difference between this and the `maker_utils` is that there is the boiler plate  involved with defining a Python `class` and manually linking that class with URIs from `rad`. There is some minor complexity with picking the correct object to inherit from, but I have detailed in the docs the process for doing this. Moreover, it is likely that any new `rad` schemas will follow a similar constructive pattern to the existing ones meaning that one should be able to follow what is laid out by these changes to create and modify new things.

I know this looks like a lot of changes, but it is mostly packing boiler plate and better organization around what already exists in the `maker_utils` currently.

Note that as far as functionality is concerned, all of these explicit objects should just act like the dynamically created subclasses of `DNode` that we already work with and that most of the "stuff" (the `@rad.field` things) in the class are effectively window dressings added on top of the `DNode`.

There are only two deviations from how RDM operated previously, neither of which are truly necessary to these changes, but I believe they are useful enhancements:

1. The `DataModel` (e.g. `ImageModel`) inherits directly from the node it was previously wrapping. This has been done mainly so that we don't have to add explicit pass throughs to the `@rad.fields` in the DataModel object in order to maintain the explicit, statically analyzable state of the model. Otherwise you end up with `__getattr__` based pass throughs which block static analysis tools. Note however, that it is still the case that only the node object is being handled by ASDF rather than the over arching DataModel object.
2. A tiny amount of "validation-like-behavior" on value setting has been re-introduced. The scope of this is limited to only the explicitly enumerated values. It is not validation in the ASDF/schema sense, but rather explicit Python `Enum` objects have been introduced which explicitly enumerate the values indicated by the schemas. The mechanisms inside `DNode` will automatically wrap values into their correct `stnode`-wrapper if one is necessary, which in this case involves wrapping the set value into the `Enum` object. That object will raise an error if the value presented is not part of the `Enum`. While this is not strictly necessary, it does mean that we do not need to keep using "magic" values in our code base and can instead call into code documented values. Again this is not strictly necessary but I think it adds a layer of documentation to the code and its generated doc.

Finally, there are some nice additional features that this PR introduces.

1. Doc strings for all the "fields" and objects are all automatically generated from the title and descriptions documenting them in the requisite `rad` schema. This is done dynamically at run-time but I made it lazy because it significantly slowed down the initial import of RDM. By this I mean the object will automatically open the schema and find the documentation for a given aspect, but it will only do this if one specifically requests the doc string. If no request for this is made (such as when sphinx builds the docs or when `help(<object>)` is called, then this process will not occur.
2. We can now introspect metadata from the `rad` schemas in its entire context without the need for a specific instance of a given data model or node. This is because all of the schema handling is done at the type level (methods/attributes of the type itself) rather than on an instance, like how `asdf_info` has to work. In particular, I have added things like `archive_catalog` metadata pulling routines that can be called on a type itself rather than on an instance of that type.
3. Save for the `maker_utils`, this PR makes RDM fully statically type checkable. Meaning that MyPy can be used to fully inspect all of RDM (meaning that it is fully self consistent aside from a small subset of things I have had MyPy ignore because fixing/making MyPy understand the logic was going to be more complex than it was worth). This means that IDEs and other tools can use type hints to assist users with their interactions with RDM now.
4. As already stated, the nodes are lazy meaning that RDM will only fill in data when it is called for (either with a default when the user requests data that it does not have, or when they explicitly set it). 

NOTE: I have a branch of `romancal` which strips it of its use of the `maker_utils` some of which were being used in ways I did not expect and broke when I linked them directly to the lazy models. This will be opened in a forthcoming PR on `romancal`.

## Tasks

</details


src/roman_datamodels/stnode/core/_d_node.py
```diff
@@ -0,0 +1,424 @@
...
+
+_T = TypeVar("_T")
+
+
+class MissingFieldError(AttributeError):
+    """Error when a field is attempted to be accessed but is missing"""
+
+
+# Once we are >3.11 -> DNode[T] can replace the Generic[T] in the class definition
+class DNode(AsdfNodeMixin[_T], MutableMapping[str, _T]):
```
WilliamJamieson, line 38: This is the `DNode` object from before. I did fully strip out the last vestiges of the automatic validation. The main changes are mostly additions so that introspection works correctly.

src/roman_datamodels/stnode/core/_d_node.py
```diff
@@ -0,0 +1,424 @@
...
+            The constructor for a default. Its callable to allow for lazy instantiation.
+            This will often take the form of a lambda, that way the argument to the lambda
+            is not evaluated until the default is actually needed, saving time and memory.
+            for things like numpy arrays.
+        """
+
+        if not self._has_node(key):
+            self._set_node(key, default())
+
+        return self._pull_node(key)
```
WilliamJamieson, line 277: This method is the backbone addition to DNode that enables the lazy magic. It allows it a value to be set using a "default" function provided it does not exist.

src/roman_datamodels/stnode/core/_field.py
```diff
@@ -0,0 +1,112 @@
...
+            # This is to handle the case where we do a getattr on the class itself
+            # I don't want to confuse the normal operations with adding this to the
+            return self
+
+        # MyPy seems to have an issue with the type of obj for the `obj._get_node` call
+        # it believes it is `_D@__get__` but it should be `_D@__init__`, not sure why.
+        #
+        # The specific cast here is to let the MyPy know that the generic _get_node method
+        # will be return the correct for the descriptor.
+        return cast(_T, obj._get_node(self.default.__name__, lambda: self.default(obj)))  # type: ignore[arg-type]
```
WilliamJamieson, line 112: This is the logic that turns

```python
@rad.field
def my_field(self) -> int:
    return 42
```

Into something that acts like a `property`, but instead of evaluating the

```python
def my_field(self) -> int:
    return 42
```

function every time `instance.my_field` is accessed. Instead it wraps that into a default setting function that is passed into the `_get_node` method I previously pointed out.

Note that the `lambda: self.default(obj)` is used to ensure that the `my_field` function that is being used as a default will only be evaluated when the default is needed. This prevents relatively expensive operations like creating large `ndarray`s from being performed every time a field accessed.

Note that `cast(type, instance)` is a no-op typing call. Meaning that it is only relevant to a type checker not the python library itself. During parsing this will be turned into a simple `instance` and have an `__annotation__` appended with this information.

src/roman_datamodels/datamodels/_asdf.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        with nuke_validation():
+            # ASDF has not implemented type hints so MyPy will complain about this
+            # until they do.
+            af = AsdfFile()  # type: ignore[no-untyped-call]
+
+            af["roman"] = self.construct_node(self._data)
+            # ASDF has not implemented type hints so MyPy will complain about this
+            # until they do.
+            af.validate()  # type: ignore[no-untyped-call]
```
WilliamJamieson, line 81: I left this in because it is the existing behavior; however, this will cause the creation of all the fields that need to be filled in for a valid file. This destroys the laziness in many cases and I don't think its removal will make any significant differences.

PR #473, opened 2025-04-01
RCAL-1004 Added support for writing source catalog data models to a parquet file

### LINKED ISSUE #472: Add functionality to save source_catalog datamodels to the parquet format
Presuming the original form of the catalog is a Roman datamodel that uses an Astropy table instance to store the catalog, a method is needed to save all the relevant information (table data, and metadata) in parquet format.

perrygreenfield:
Resolves [RCAL-1004](https://jira.stsci.edu/browse/RCAL-1004)

Closes #472

This PR adds the ability to save source catalog data in I`mageSourceCatalogModel `and `MosaicSourceCatalogModel` in the parquet format. It saves the table metadata and the relevant table column metadata (primarily the unit).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,8 @@ dependencies = [
   # "rad >=0.23.1",
   "rad @ git+https://github.com/spacetelescope/rad.git",
   "asdf-standard >=1.1.0",
+  "pyarrow >=19.0.1",
+  "pyarrow.parquet"
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ dependencies = [
   "rad @ git+https://github.com/spacetelescope/rad.git",
   "asdf-standard >=1.1.0",
   "pyarrow >=19.0.1",
-  "pandas >=2.2.3",
 ]
 dynamic = ["version"]
 
@@ -34,6 +33,7 @@ test = [
   "pytest-doctestplus",
   "pytest-doctestplus >=1.2.1",
   "pytest-env >= 0.8",
+  "pandas >=2.2.3",
 ]
 docs = ["sphinx", "sphinx-automodapi", "sphinx-rtd-theme", "sphinx-astropy"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
   "rad @ git+https://github.com/spacetelescope/rad.git",
   "asdf-standard >=1.1.0",
   "pyarrow >=19.0.1",
+  "pandas >=2.2.3",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ dependencies = [
   "rad @ git+https://github.com/spacetelescope/rad.git",
   "asdf-standard >=1.1.0",
   "pyarrow >=19.0.1",
-  "pyarrow.parquet"
 ]
 dynamic = ["version"]
 
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -6,6 +6,8 @@ This module provides all the specific datamodels used by the Roman pipeline.
     from the schema manifest defined by RAD.
 """
 
+import numpy as np
+import pathlib
 import asdf
 from astropy.table import QTable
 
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -20,6 +20,8 @@ dependencies = [
   # "rad >=0.23.1",
   "rad @ git+https://github.com/spacetelescope/rad.git",
   "asdf-standard >=1.1.0",
+  "pyarrow >=19.0.1",
+  "pandas >=2.2.3",
```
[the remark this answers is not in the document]
braingram: It looks like it's a documentation issue that astropy can't read parquet files without pandas: https://github.com/astropy/astropy/issues/17972

So if we ever want to `Table.read("some.parq")` we will need to add pandas as a dependency.

tests/test_parquet.py
```diff
@@ -0,0 +1,36 @@
...
+
+@pytest.mark.parametrize(("catalog_class", "mk_catalog"), source_catalogs)
+def test_source_catalog(catalog_class, mk_catalog):
+    sc_node = mk_catalog(save=False)
+    sc_dm = catalog_class(sc_node)
+    bio = io.BytesIO()
+    sc_dm.to_parquet(file=bio)
+    bio.seek(0)
+
+    ptab = astrotab.Table.read(bio, format="parquet")
```
braingram, line 25:
```suggestion
    ptab = pq.ParquetFile(bio).read().to_pydict()
```
Does this work to allow this PR to avoid adding pandas as a dependency?

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -6,16 +6,97 @@
...
 
 __all__ = []
 
+DTYPE_MAP = {}
+
+
+class _ParquetMixin:
+    """Gives SourceCatalogModels the ability to save to parquet files."""
+
+    def to_parquet(self, file=None, fdir=None):
```
braingram, line 29:
```suggestion
    def to_parquet(self, filepath):
```
Can we simplify this a bit to just accept a path?

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -6,16 +6,97 @@
...
+            else:
+                filename_root = pathlib.Path(self.meta.filename[:-5])
+            filename_root = filename_root.with_suffix(".parquet")
+            if fdir is not None:
+                filepath = pathlib.Path(fdir) / filename_root
+            else:
+                filepath = pathlib.Path(".") / filename_root
+        # Otherwise assume it is a file-like object
+        else:
+            filepath = file
```
braingram, line 72:
```suggestion
```

tests/test_parquet.py
```diff
@@ -0,0 +1,36 @@
...
+from roman_datamodels import maker_utils as utils
+
+source_catalogs = [
+    (datamodels.ImageSourceCatalogModel, utils.mk_image_source_catalog),
+    (datamodels.MosaicSourceCatalogModel, utils.mk_mosaic_source_catalog),
+]
+
+
+@pytest.mark.parametrize(("catalog_class", "mk_catalog"), source_catalogs)
+def test_source_catalog(catalog_class, mk_catalog):
```
braingram, line 18:
```suggestion
def test_source_catalog(catalog_class, mk_catalog, tmp_path):
```

tests/test_parquet.py
```diff
@@ -0,0 +1,36 @@
...
+    sc_node = mk_catalog(save=False)
+    sc_dm = catalog_class(sc_node)
+    bio = io.BytesIO()
+    sc_dm.to_parquet(file=bio)
```
braingram, line 22:
```suggestion
    test_path = tmp_path / "test.parquet"
    sc_dm.to_parquet(test_path)
```

tests/test_parquet.py
braingram, line 25:
```suggestion
    ptab = astrotab.Table.read(test_path, format="parquet")
```

tests/test_parquet.py
```diff
@@ -0,0 +1,36 @@
+import io
+
```
braingram, line 2:
```suggestion
```

braingram (commented):
Thanks for making the pandas dependency update. I left a few comments about simplifying the input handling. If the plan is to use this in the pipeline is there ever a time the pipeline will provide something other than a filepath or string? I can't think of one.

nden (approved):

src/roman_datamodels/datamodels/_datamodels.py
perrygreenfield: I think so. I'll do that now.

tests/test_parquet.py
perrygreenfield: Done.

tests/test_parquet.py
perrygreenfield: Done.

tests/test_parquet.py
perrygreenfield: Done.

braingram (approved):
Thanks! LGTM.

braingram:
As far as I'm aware parquet does not support nested metadata and requires that "strings" be byte encoded. Is there documentation that says otherwise?

larrybradley:
Thanks.  I haven't read the parquet standard, but I was surprised when I saw how the metadata were formatted.  Are the parquet files written by Astropy Table (via pandas) non-standard?  Those work as expected:

```python
>>> from astropy.table import QTable
>>> t = QTable()
>>> t['col1'] = [1, 2, 3]
>>> t['col2'] = ['a', 'b', 'c']
>>> t['col1'].unit = 'nJy'
>>> t['col1'].description = 'some integers'
>>> t['col2'].description = 'some letters'
>>> t.meta['telescope'] = 'Roman'
>>> t.meta['instrument'] = 'WFI'
>>> t.meta['filter'] = 'F087'
>>> phot = {"conversion_megajanskys": -999999, "conversion_megajanskys_uncertainty": -999999, "pixel_area": -999999}
>>> t.meta['photometry'] = phot
>>> fn = 'test.parquet'
>>> t.write(fn, overwrite=True)
>>> tbl = QTable.read(fn)
>>> print(tbl)
col1 col2
nJy      
---- ----
 1.0    a
 2.0    b
 3.0    c


>>> tbl.meta
{'telescope': 'Roman',
 'instrument': 'WFI',
 'filter': 'F087',
 'photometry': {'conversion_megajanskys': -999999,
  'conversion_megajanskys_uncertainty': -999999,
  'pixel_area': -999999}}

>>> tbl['col1'].unit
nJy

>>> tbl['col1'].description
'some integers'
```

I have a few concerns about the current parquet metadata:

* Typical astronomers are going to use Astropy Table or pandas to read the files, and it appears the table and column metadata (e.g., units) are not included when doing so (please let me know if I'm missing something).  Is it expected that users will need to use the separate pyarrow `read_schema` function to get the table metadata?  I don't think that's going to be obvious or user friendly -- I have a feeling that the table metadata that we generate will effectively be invisible to most astronomers. Is there a better way?

```python
>>> import pyarrow.parquet as pq
>>> tbl_schema = pq.read_schema('test.parquet')
>>> tbl_schema.field("x_centroid").metadata
{b'unit': b'pix'}
```

* The original column descriptions (e.g., `tbl[col].description`) are currently dropped in these parquet files.  Currently, I explicitly add column descriptions in the output source catalog to help make the catalog files self documenting.  Are there any plans to support that?   If not, I'll remove them from the source catalog code.

* Retrieving the table metadata value in the original data type requires some gymnastics that isn't going to be obvious to most astronomers:  Is there a better way than, e.g., this?
```python
>>> radii = np.fromstring(tbl_schema.metadata[b'source_catalog.aperture_radii.circle_pix'].decode('utf-8').strip('[]'), sep=' ')
array([ 1.00000007,  2.00000014,  4.00000028,  8.00000056, 16.00000113])
```

Are there plans to enable reading parquet files back into the source catalog datamodels (via an "open" method)?

CC: @schlafly

braingram:
> Are the parquet files written by Astropy Table (via pandas) non-standard? 

Is there a standard that defines how the astropy Table metadata is defined? As far as I know there is not.

If I write a table with metadata:
```python
t.meta['a'] = {'b': {'c': 1}}
```
It produces a file with no file-level metadata and the following schema metadata:
```python
{b'table_meta_yaml': b'datatype:\n- {name: a, datatype: int64}\nmeta: !!omap\n- a:\n    b: {c: 1}'}
```
which is a single key-value pair with key being `b'table_meta_yaml'` and the value being some yaml formatted bytes.

> Typical astronomers are going to use Astropy Table or pandas to read the files, and it appears the table and column metadata (e.g., units) are not included when doing so (please let me know if I'm missing something). Is it expected that users will need to use the separate pyarrow read_schema function to get the table metadata? I don't think that's going to be obvious or user friendly -- I have a feeling that the table metadata that we generate will effectively be invisible to most astronomers. Is there a better way?

Are column units part of parquet or something specific to astropy tables? As far as I can tell there is no standard "unit" definition for parquet. If I set a unit for the example `a` column above astropy stores the unit information in the yaml encoded bytes: 
```python
{b'table_meta_yaml': b'datatype:\n- {name: a, unit: deg, datatype: int64}\nmeta: !!omap\n- a:\n    b: {c: 1}'}
```

Do you have a suggestion for where to move this discussion? I'm not sure a closed PR has enough visibility.

schlafly:
I have a PR here where we can continue this discussion: https://github.com/spacetelescope/roman_datamodels/pull/488 .  That adds the table_meta_yaml following your ~suggestion so that astropy knows about units and descriptions.  It's obviously kind of garbage in that it duplicates a bunch of metadata in a long annoying yaml string, but it's nice that it makes astropy happy.

PR #477, opened 2025-04-03
RCAL-1025 Create the WfiWcsModel

stscieisenhamer:
WIP [RCAL-1015](https://jira.stsci.edu/browse/RCAL-1025)
Depends on[ RAD#564](https://github.com/spacetelescope/rad/pull/564)

This PR creates the datamodel `WfiWcsModel` based on the schema `WfiWcs`. This model will be used to save the L2 wcs information.

## Tasks
- [x] Update or add relevant `roman_datamodels` tests.
- [x] Update relevant docstrings and / or `docs/` page.
- [x] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [x] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [x] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

</details

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
   "astropy >=5.3.0",
   # "rad >=0.23.1",
   #  "rad @ git+https://github.com/spacetelescope/rad.git",
-  "rcal-1025-wcsasdf @ git+https://github.com/stscieisenhamer/rad.git",
+  "rad @ git+https://github.com/stscieisenhamer/rad.git@rcal-1025-wcsasdf",
   "asdf-standard >=1.1.0",
   "pyarrow >=19.0.1",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,8 +18,7 @@ dependencies = [
   "numpy >=1.24",
   "astropy >=5.3.0",
   # "rad >=0.23.1",
-  #  "rad @ git+https://github.com/spacetelescope/rad.git",
-  "rad @ git+https://github.com/stscieisenhamer/rad.git@rcal-1025-wcsasdf",
+  "rad @ git+https://github.com/spacetelescope/rad.git",
   "asdf-standard >=1.1.0",
   "pyarrow >=19.0.1",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
   "astropy >=5.3.0",
   # "rad >=0.23.1",
   #  "rad @ git+https://github.com/spacetelescope/rad.git",
-  "git+https://github.com/stscieisenhamer/rad@rcal-1025-wcsasdf",
+  "rcal-1025-wcsasdf @ git+https://github.com/stscieisenhamer/rad.git",
   "asdf-standard >=1.1.0",
   "pyarrow >=19.0.1",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,8 @@ dependencies = [
   "numpy >=1.24",
   "astropy >=5.3.0",
   # "rad >=0.23.1",
-  "rad @ git+https://github.com/spacetelescope/rad.git",
+  #  "rad @ git+https://github.com/spacetelescope/rad.git",
+  "git+https://github.com/stscieisenhamer/rad@rcal-1025-wcsasdf",
   "asdf-standard >=1.1.0",
   "pyarrow >=19.0.1",
 ]
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -5,14 +5,14 @@ This module provides all the specific datamodels used by the Roman pipeline.
     the top-level STNode type that the datamodel wraps. This STNode type is derived
     from the schema manifest defined by RAD.
 """
+import copy
 
 import asdf
 import numpy as np
+from astropy.modeling import models
 from astropy.table import QTable
 
-from roman_datamodels import stnode
-
-from ..stnode import DNode
+from .. import stnode
 from ._core import DataModel
 from ._utils import _node_update
 
@@ -60,7 +60,7 @@ class _ParquetMixin:
         source_cat = self.source_catalog
         scmeta = source_cat.meta
         # Wrap it as a DNode so it can be flattened
-        dn_scmeta = DNode(scmeta)
+        dn_scmeta = stnode.DNode(scmeta)
         flat_scmeta = dn_scmeta.to_flat_dict(recursive=True)
         # Add prefix to flattened keys to indicate table metadata
         flat_scmeta = {"source_catalog." + k: str(v) for (k, v) in flat_scmeta.items()}
@@ -446,7 +446,7 @@ class WfiWcsModel(_RomanDataModel):
     _node_type = stnode.WfiWcs
 
     @classmethod
-    def from_model_with_wcs(cls, model):
+    def from_model_with_wcs(cls, model, l1_border=4):
         """Extract the WCS information from an exposure model post-assign_wcs
 
         Construct a `WfiWcsModel` from any model that is used post-assign_wcs step
@@ -461,6 +461,9 @@ class WfiWcsModel(_RomanDataModel):
         ----------
         model : ImageModel
             The input data model (a WfiWcsModel will also work).
+
+        l1_border : int
+            The extra border to add for the L1 wcs.
 
         Returns
         -------
@@ -485,6 +488,15 @@ class WfiWcsModel(_RomanDataModel):
 
         # Assign the model WCS to the L2-specified wcs attribute
         wfi_wcs['wcs_l2'] = model.meta.wcs
+
+        # Create an L1 WCS that accounts for the extra border.
+        l1_wcs = copy.deepcopy(model.meta.wcs)
+        l1_shift = models.Shift(-l1_border) & models.Shift(-l1_border)
+        l1_wcs.insert_transform('detector', l1_shift, after=True)
+        bb = wfi_wcs['wcs_l2'].bounding_box
+        if bb is not None:
+            l1_wcs.bounding_box = ((bb[0][0], bb[0][1] + 2 * l1_border), (bb[1][0], bb[1][1] + 2 * l1_border))
+        wfi_wcs['wcs_l1'] = l1_wcs
 
         # Create model from node
         wfi_wcs_model = WfiWcsModel(wfi_wcs)
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -481,7 +481,7 @@ class WfiWcsModel(_RomanDataModel):
 
         wfi_wcs = mk_wfi_wcs()
 
-        _node_update(wfi_wcs, model)
+        _node_update(wfi_wcs, model, only_in_to_node=True)
 
         # Create model from node
         wfi_wcs_model = WfiWcsModel(wfi_wcs)
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -478,28 +478,25 @@ class WfiWcsModel(_RomanDataModel):
             a WfiWcsModel, it is simply returned.
 
         """
-        ALLOWED_MODELS = ImageModel
-
         if isinstance(model, cls):
             return model
-        if not isinstance(model, ALLOWED_MODELS):
-            raise ValueError(f"Input must be one of {ALLOWED_MODELS}")
+        if not isinstance(model, ImageModel):
+            raise ValueError("Input must be an ImageModel")
 
         # Retrieve the needed meta components
         wfi_wcs = cls()
         wfi_wcs.meta = {}
-        skip_keys = {}
         for k in wfi_wcs.meta._schema_attributes.explicit_properties:
-            if k not in skip_keys and k in model.meta:
+            if k in model.meta:
                 wfi_wcs.meta[k] = copy.deepcopy(model.meta[k])
 
-        # Ensure a WCS has been defined.
+        # Check that a WCS has been defined.
         if model.meta.wcs is None:
             log.info("Model has no WCS defined. Will not populate the WCS components.")
             return wfi_wcs
 
         # Assign the model WCS to the L2-specified wcs attribute
-        wfi_wcs.wcs_l2 = model.meta.wcs
+        wfi_wcs.wcs_l2 = copy.deepcopy(model.meta.wcs)
 
         # Create an L1 WCS that accounts for the extra border.
         l1_wcs = copy.deepcopy(model.meta.wcs)
@@ -511,10 +508,8 @@ class WfiWcsModel(_RomanDataModel):
         wfi_wcs.wcs_l1 = l1_wcs
 
         # Get alignment results, if available
-        try:
-            wfi_wcs.meta.wcs_fit_results = model.meta.wcs_fit_results.value
-        except AttributeError:
-            pass
+        if hasattr(model.meta, 'wcs_fit_results'):
+            wfi_wcs.meta.wcs_fit_results = copy.deepcopy(model.meta.wcs_fit_results)
 
         # That's all folks.
         return wfi_wcs
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -459,7 +459,7 @@ class WfiWcsModel(_RomanDataModel):
 
         Parameters
         ----------
-        model : RampModel, ImageModel
+        model : ImageModel
             The input data model (a WfiWcsModel will also work).
 
         Returns
@@ -469,7 +469,7 @@ class WfiWcsModel(_RomanDataModel):
             a WfiWcsModel, it is simply returned.
 
         """
-        ALLOWED_MODELS = (RampModel, ImageModel)
+        ALLOWED_MODELS = (ImageModel)
 
         if isinstance(model, cls):
             return model
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -506,7 +506,7 @@ class WfiWcsModel(_RomanDataModel):
 
         # Get alignment results, if available
         if hasattr(model.meta, "wcs_fit_results"):
-            wfi_wcs.meta.wcs_fit_results = copy.deepcopy(model.meta.__getitem__("wcs_fit_results"))
+            wfi_wcs.meta.wcs_fit_results = copy.deepcopy(model.meta["wcs_fit_results"])
 
         # That's all folks.
         return wfi_wcs
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -440,3 +440,7 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
 
 class SegmentationMapModel(_RomanDataModel):
     _node_type = stnode.SegmentationMap
+
+
+class WfiWcsModel(_RomanDataModel):
+    _node_type = stnode.WfiWcs
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -5,6 +5,7 @@ This module provides all the specific datamodels used by the Roman pipeline.
     the top-level STNode type that the datamodel wraps. This STNode type is derived
     from the schema manifest defined by RAD.
 """
+
 import copy
 
 import asdf
@@ -472,7 +473,7 @@ class WfiWcsModel(_RomanDataModel):
             a WfiWcsModel, it is simply returned.
 
         """
-        ALLOWED_MODELS = (ImageModel)
+        ALLOWED_MODELS = ImageModel
 
         if isinstance(model, cls):
             return model
@@ -481,7 +482,7 @@ class WfiWcsModel(_RomanDataModel):
 
         # Ensure a WCS has been defined.
         if model.meta.wcs is None:
-            raise ValueError(f'Model has no WCS defined, cannot create {cls}')
+            raise ValueError(f"Model has no WCS defined, cannot create {cls}")
 
         # Create base node with dummy values (for validation)
         from roman_datamodels.maker_utils import mk_wfi_wcs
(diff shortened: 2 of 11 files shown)


src/roman_datamodels/testing.py
```diff
@@ -109,13 +109,18 @@ def _assert_value_is_copy(value1, value2, deepcopy=False):
...
-            # Hashable objects cannot be copied as they are immutable.
+            # If not hashable, then can be copied.
             try:
                 hash(value1)
             except TypeError:
                 assert value1 is not value2
             else:
-                assert value1 is value2
+                # Normally hashable objects do not deepcopy.
+                # GWCS violates this.
```
schlafly, line 119: Just out of curiosity---I would have said we already had models with GWCS in them (e.g., WfiImage).  What's different here?

stscieisenhamer: On my list to investigate. The known difference is that the GWCS objects are usually in `meta.wcs` and not top-level, such as these are. This implies that a different decision path occurs WRT the meta tree.

Regardless, the code itself is based on a false assumption; [see python docs, section 6.10.1](https://docs.python.org/3/reference/expressions.html#value-comparisons). The relevant words are at the end of that section about `hash()` and equality. Python does not enforce/cannot enforce this, since a class can define its own hash. As such, the current code modification is theoretically needed anyways. Regarding GWCS itself, an issue will be created for discussion there.

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +441,74 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+        # Ensure a WCS has been defined.
+        if model.meta.wcs is None:
+            raise ValueError(f"Model has no WCS defined, cannot create {cls}")
+
+        # Create base node with dummy values (for validation)
+        from roman_datamodels.maker_utils import mk_wfi_wcs
+
+        wfi_wcs = mk_wfi_wcs()
+
+        _node_update(wfi_wcs, model, only_in_to_node=True)
```
schlafly, line 492: FYI, Brett has been trying to eliminate uses of the maker utilities in the pipeline, preferring to directly create the model and then to fill out the metadata.  e.g., https://github.com/spacetelescope/romancal/pull/1675/files  The rationale is that this leave less room to forget to pass through metadata that needs to be passed through.  Does that approach work here?

tests/test_models.py
```diff
@@ -1197,3 +1198,93 @@ def test_apcorr_none_array():
...
+            assert_node_equal(wfi_wcs_value, model_value)
+
+        else:
+            raise ValueError(f"Unexpected type {type(wfi_wcs_value)}, {key}")  # pragma: no cover
+
+    # Test wcs fidelity
+    border = 4.0  # Default extra border for L1
+    model_corner = model.meta.wcs.pixel_to_world(0.0, 0.0)
+    wfi_wcs_corner = wfi_wcs.wcs_l1.pixel_to_world(border, border)  # Extra border due to being L1
+    assert model_corner.separation(wfi_wcs_corner).value <= 1e-5
```
schlafly, line 1276: This is the test I most wanted to see.  I'm a little afraid of the 1e-5 though; what precision is actually delivered?  1e-5 degrees is 0.036" is ~one third of a pixel; I hope we do ~100x better than this.

src/roman_datamodels/datamodels/_datamodels.py
stscieisenhamer: I had asked about this in the slack channel. I believe the answer is no because the use-case is a bit different. So far all the conversion methods are meant to transfer information between two, possibly very different, schemas. Hence the attributes are not necessarily one-to-one and depend on knowing what the trees are. At the moment, the only easy way of knowing what the tree looks like are the maker utils. 

Williams' and others work on the model refactoring should deal with this in a cleaner way, but that is, I believe, the only path we have.

tests/test_models.py
stscieisenhamer: I just pulled that out-of-the-hat; I believe the current actual difference is 0 but wanted to account for some level of numerical precision issue, like usually. I can bring it down to a more precision value such as 1.e-7.

src/roman_datamodels/datamodels/_datamodels.py
braingram, line 492:
```suggestion
        wfi_wcs = cls()
        wfi_wcs.meta = {}
        skip_keys = {"calibration_software_version", "product_type", "filename", "file_date"}
        for k in wfi_wcs.meta._schema_attributes.explicit_properties:
            if k not in skip_keys and k in model.meta:
                wfi_wcs.meta[k] = copy.deepcopy(model.meta[k])
```
Generally yes there's no substitute for the maker_utils. In this case the above suggestion would copy over the schema-defined metadata from the input model (without introducing new fake values). It might be easier at this point to remove the `skip_keys` although it does highlight that having this conversion code in roman_datamodels doesn't allow access to the romancal version. Any reason to not have this conversion code in romancal?

Also, what should `product_type` be for this model?

Finally, as there's no real solution for `file_date` copying the junk value from the input seems like a reasonable option here. Perhaps the same is true for `filename` although `none` seems marginally better.

stscieisenhamer:
[Regression Test #1](https://github.com/spacetelescope/RegressionTests/actions/runs/14334781965)

src/roman_datamodels/datamodels/_datamodels.py
stscieisenhamer: Concerning the statement "conversion code in roman_datamodels doesn't allow access to the romancal version": What is meant by "version"? This is simply a conversion of one model to another and not sure what dependency on the any higher level package should be, whether it is romancal or something else?

But this does answer the question asked in slack about "is there a way of getting the tree", and it appears there is. I will explore that. However, this will involve updating of `_node_update` because there is other functionality needed. But if that can be traversed using the `_schema_attributes`, we'll take it.

stscieisenhamer: I see, the `_schema_attributes` only exists for `meta`. So top level attributes cannot be devined from this way. For this case, the above will work. Thanks!

braingram: > Concerning the statement "conversion code in roman_datamodels doesn't allow access to the romancal version": What is meant by "version"? This is simply a conversion of one model to another and not sure what dependency on the any higher level package should be, whether it is romancal or something else?

"meta.calibration_software_version" records the romancal version number/string. This is the version I was referring to. It's required to be defined for a valid file and is typically set by `Step.finalize_result` in romancal. The point I was trying to make was that if this conversion code was in romancal it could fill in that version string.

> But this does answer the question asked in slack about "is there a way of getting the tree", and it appears there is. I will explore that. However, this will involve updating of `_node_update` because there is other functionality needed. But if that can be traversed using the `_schema_attributes`, we'll take it.
>
> I see, the _schema_attributes only exists for meta. So top level attributes cannot be devined from this way. For this case, the above will work. Thanks!

`_schema_attributes` exists for any `DNode`: `meta`, also `meta.exposure`, etc and even the top level node `WfiWcsNode` (or something like that) in this case. This doesn't fully define the "tree" as the schema can contain things like "patternProperties" which constrain a possible value but don't define an "explicit" value.

stscieisenhamer: will explore that bit later, but now just using the provided code.

src/roman_datamodels/testing.py
```diff
@@ -109,13 +109,18 @@ def _assert_value_is_copy(value1, value2, deepcopy=False):
...
+                if isinstance(value1, gwcs.WCS) and isinstance(value2, gwcs.WCS):
+                    assert True
```
braingram, line 121:
```suggestion
                    assert value1 is not value2
```
Admittedly this is not a fix for the issue you mentioned in https://github.com/spacetelescope/roman_datamodels/pull/477/files#r2027537153 but at least this will catch the case where both wcses are the same object (and not a copy).

tests/test_models.py
```diff
@@ -1197,3 +1198,100 @@ def test_apcorr_none_array():
...
+                elif meta_key == "cal_step":
+                    continue
+                if meta_key in model_meta:
+                    assert wfi_wcs_meta[meta_key] == model_meta[meta_key]
+
+        elif isinstance(wfi_wcs_value, stnode.DNode):
+            assert_node_equal(wfi_wcs_value, model_value)
+
+        else:
+            raise ValueError(f"Unexpected type {type(wfi_wcs_value)}, {key}")  # pragma: no cover
```
braingram, line 1270:
```suggestion
    for key in wfi_wcs.meta:
        wfi_wcs_value = getattr(wfi_wcs.meta, key)
        model_value = getattr(model.meta, key)
        assert_node_equal(wfi_wcs_value, model_value)
```
Since this doesn't test `wcs_l1` or `wcs_l2` (they don't exist in `model`) this block is only testing `meta` (see the lack of coverage of the test). It can be simplified to the above suggestion.

braingram (commented):
I added a few comments. Is the RAD branch up-to-date? It requires `wcs_l1` and `wcs_l2` yet the conversion utility here doesn't populate them if `model.meta.wcs` isn't defined and instead returns a known invalid `WfiWcsModel`.

stscieisenhamer:
Concerning the requirements on the wcs properties: That will be removed.

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +446,75 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+        -------
+        wfiwcs_model : WfiWcsModel
+            The WfiWcsModel built from the input model. If the input is already
+            a WfiWcsModel, it is simply returned.
+
+        """
+        ALLOWED_MODELS = ImageModel
+
+        if isinstance(model, cls):
+            return model
```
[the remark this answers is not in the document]
stscieisenhamer: Yes, that is the name of the model.

Is the intent of removal of this line to have the method make a new copy?

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +446,75 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+        if not isinstance(model, ALLOWED_MODELS):
+            raise ValueError(f"Input must be one of {ALLOWED_MODELS}")
```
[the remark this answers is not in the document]
stscieisenhamer: Because the above just returns the input.

Otherwise, agreed.

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +446,75 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+        wfi_wcs.meta = {}
+        skip_keys = {}
+        for k in wfi_wcs.meta._schema_attributes.explicit_properties:
+            if k not in skip_keys and k in model.meta:
+                wfi_wcs.meta[k] = copy.deepcopy(model.meta[k])
+
+        # Ensure a WCS has been defined.
+        if model.meta.wcs is None:
+            log.info("Model has no WCS defined. Will not populate the WCS components.")
+            return wfi_wcs
```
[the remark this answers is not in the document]
stscieisenhamer: RAD will be updated.

src/roman_datamodels/datamodels/_datamodels.py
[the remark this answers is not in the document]
braingram: The current romancal PR only ever calls this method with an `ImageModel` instance:
https://github.com/spacetelescope/romancal/pull/1680/files#diff-eb70480fc65aab38ba627de35bc7b5081bf08f4c4cee6d4b98bf3445df7e9a12R24

When would calling this with a `WfiWcsModel` be used or useful?

tests/test_models.py
stscieisenhamer: It does have a wcs. Breaking at #1218 gives:
```
(Pdb) p model.meta.wcs
<WCS(output_frame=icrs, input_frame=detector, forward_transform=Model: CompoundModel
Inputs: ('x0', 'x1')
Outputs: ('alpha_C', 'delta_C')
Model set size: 1
Expression: [0] & [1] | [2] & [3] | [4] | [5]
Components:
    [0]: <Shift(offset=-500.)>

    [1]: <Shift(offset=-500.)>

    [2]: <Scale(factor=0.00002778)>

    [3]: <Scale(factor=0.00002778)>

    [4]: <Pix2Sky_Gnomonic()>

    [5]: <RotateNative2Celestial(lon=30., lat=45., lon_pole=180.)>
Parameters:
    offset_0 offset_1        factor_2              factor_3             lon_5        lat_5 lon_pole_5
    -------- -------- --------------------- --------------------- ------------------ ----- ----------
      -500.0   -500.0 2.777777777777778e-05 2.777777777777778e-05 29.999999999999996  45.0      180.0)>
```

and after line 1240 which is
```
    wfi_wcs = datamodels.WfiWcsModel.from_model_with_wcs(model)
```
gives `wfi_wcs`:
```
(Pdb) p wfi_wcs.wcs_l1
<WCS(output_frame=icrs, input_frame=detector, forward_transform=Model: CompoundModel
Inputs: ('x0', 'x1')
Outputs: ('alpha_C', 'delta_C')
Model set size: 1
Expression: [0] & [1] | [2] & [3] | [4] & [5] | [6] | [7]
Components:
    [0]: <Shift(offset=-4.)>

    [1]: <Shift(offset=-4.)>

    [2]: <Shift(offset=-500.)>

    [3]: <Shift(offset=-500.)>

    [4]: <Scale(factor=0.00002778)>

    [5]: <Scale(factor=0.00002778)>

    [6]: <Pix2Sky_Gnomonic()>

    [7]: <RotateNative2Celestial(lon=30., lat=45., lon_pole=180.)>
Parameters:
    offset_0 offset_1 offset_2 offset_3        factor_4              factor_5             lon_7        lat_7 lon_pole_7
    -------- -------- -------- -------- --------------------- --------------------- ------------------ ----- ----------
        -4.0     -4.0   -500.0   -500.0 2.777777777777778e-05 2.777777777777778e-05 29.999999999999996  45.0      180.0)>
```

stscieisenhamer: Oh, but I see what you are getting at. Will include the test of the wcs's also.

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +446,68 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+            The input data model (a WfiWcsModel will also work).
+
+        l1_border : int
+            The extra border to add for the L1 wcs.
+
+        Returns
+        -------
+        wfiwcs_model : WfiWcsModel
+            The WfiWcsModel built from the input model. If the input is already
+            a WfiWcsModel, it is simply returned.
```
braingram, line 478:
```suggestion
            The WfiWcsModel built from the input model.
```

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +446,68 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+        The wcs-related meta information is copied verbatim from the input model.
+
+        However, the WCS object itself is placed into the attribute 'wcs_l2'. Furthermore, a
+        modified GWCS, applicable to the Level 1 version of the input model, is created
+        and stored in the attribute 'wcs_l1'.
+
+        Parameters
+        ----------
+        model : ImageModel
+            The input data model (a WfiWcsModel will also work).
```
braingram, line 469:
```suggestion
            The input data model.
```

pyproject.toml
```diff
@@ -18,7 +18,8 @@ dependencies = [
   "numpy >=1.24",
   "astropy >=5.3.0",
   # "rad >=0.23.1",
-  "rad @ git+https://github.com/spacetelescope/rad.git",
+  #  "rad @ git+https://github.com/spacetelescope/rad.git",
+  "rad @ git+https://github.com/stscieisenhamer/rad.git@rcal-1025-wcsasdf",
```
braingram, line 22: Reminder comment to update this once the rad PR is merged.

src/roman_datamodels/testing.py
braingram: Would you open a roman_datamodels issue or rcal ticket about this?

src/roman_datamodels/testing.py
braingram: I'm not seeing this suggestion applied. Was it causing an issue?

tests/test_models.py
```diff
@@ -1197,3 +1198,77 @@ def test_apcorr_none_array():
...
+    assert model_bb[0][1] + 2 * border == wfi_wcs_bb[0][1]
+    assert model_bb[1][1] + 2 * border == wfi_wcs_bb[1][1]
+
+
+def test_wfi_wcs_no_wcs(caplog):
+    """Test that file is still generate with lack of wcs"""
+    model = datamodels.ImageModel(utils.mk_level2_image(shape=(8, 8)))
+    model.meta.wcs = None
+
+    wfi_wcs = datamodels.WfiWcsModel.from_model_with_wcs(model)
```
braingram, line 1266:
```suggestion
    wfi_wcs = datamodels.WfiWcsModel.from_model_with_wcs(model)
    assert wfi_wcs.validate() is None
```

tests/test_models.py
```diff
@@ -1197,3 +1198,77 @@ def test_apcorr_none_array():
...
+
+    with pytest.raises(AttributeError):
+        _ = wfi_wcs.wcs_l1
+
+    with pytest.raises(AttributeError):
+        _ = wfi_wcs.wcs_l2
```
braingram, line 1272:
```suggestion
    assert not hasattr(wfi_wcs, "wcs_l1")
    
    assert not hasattr(wfi_wcs, "wcs_l2")
```

tests/test_models.py
```diff
@@ -1197,3 +1198,77 @@ def test_apcorr_none_array():
...
+    assert model_corner.separation(wfi_wcs_corner).value <= 1e-7
+
+    model_bb = model.meta.wcs.bounding_box
+    wfi_wcs_bb = wfi_wcs.wcs_l1.bounding_box
+    assert model_bb[0][1] + 2 * border == wfi_wcs_bb[0][1]
+    assert model_bb[1][1] + 2 * border == wfi_wcs_bb[1][1]
+
+
+def test_wfi_wcs_no_wcs(caplog):
+    """Test that file is still generate with lack of wcs"""
```
braingram, line 1262:
```suggestion
    """Test providing an ImageModel without a wcs produces a valid WfiWcsModel with no wcses."""
```

tests/test_models.py
```diff
@@ -1197,3 +1198,77 @@ def test_apcorr_none_array():
...
+        "rot": [1.2078100852299566e-05, 1.2078100852299566e-05],
+        "scale": [1.0, 1.0],
+        "shift": [0.00017039070698617517, -0.00023752675967125825],
+        "skew": 0.0,
+        "status": "SUCCESS",
+    }
+
+    wfi_wcs = datamodels.WfiWcsModel.from_model_with_wcs(model)
+
+    # Test for equality of attributes
```
braingram, line 1243:
```suggestion
    # Test meta copied from input model
```

braingram (commented):
Thanks for making these updates. The main code looks good to me. I left a few more docs and testing comments.

braingram:
Interesting. Attempting to call `validate` on a newly created model (as is done in `test_wfi_wcs_no_wcs`) causes an exception. That issue wasn't introduced by this PR as doing the following on main is sufficient to produce the error:
```python
m = rdm.ImageModel()
m.validate()
```
produces:
```
File ~/projects/src/roman_datamodels/src/roman_datamodels/datamodels/_core.py:350, in DataModel.validate(self)
    346 def validate(self):
    347     """
    348     Re-validate the model instance against the tags
    349     """
--> 350     self._asdf.validate()

AttributeError: 'NoneType' object has no attribute 'validate'
```

One option would be to fix this bug as part of this PR. Testing locally a single line change is needed. Adding `@_set_default_asdf` before `def validate` here fixes the issue (and doesn't cause any other tests to fail):
https://github.com/spacetelescope/roman_datamodels/blob/f8ea3b8de8ad839e50472f2558df2ad9839705bb/src/roman_datamodels/datamodels/_core.py#L345-L350

The other option would be to remove the `validate` call added in https://github.com/spacetelescope/roman_datamodels/pull/477/commits/541e5175f47e66bc56e06d4dbb88b27bade5b0fa although that would make the test less useful (an invalid model could be returned).

@stscieisenhamer any objection to making the one line bugfix above adding the `_set_default_asdf` decorator to `validate`?

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +446,67 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+        l1_shift = models.Shift(-l1_border) & models.Shift(-l1_border)
+        l1_wcs.insert_transform("detector", l1_shift, after=True)
+        bb = wfi_wcs["wcs_l2"].bounding_box
+        if bb is not None:
+            l1_wcs.bounding_box = ((bb[0][0], bb[0][1] + 2 * l1_border), (bb[1][0], bb[1][1] + 2 * l1_border))
+        wfi_wcs.wcs_l1 = l1_wcs
+
+        # Get alignment results, if available
+        if hasattr(model.meta, "wcs_fit_results"):
+            wfi_wcs.meta.wcs_fit_results = copy.deepcopy(model.meta.wcs_fit_results)
```
braingram, line 509:
```suggestion
        if hasattr(model.meta, "wcs_fit_results"):
            wfi_wcs.meta.wcs_fit_results = copy.deepcopy(model.meta.wcs_fit_results)
```
I just tried to run roman_elp with https://github.com/spacetelescope/romancal/pull/1680/ and this PR. It crashed on this line with `TypeError: cannot pickle 'dict_keys' object`. I don't know what's going on there but changing this to `__getitem__` seems to fix it.
I will try to come up with a minimum reproducer and open an issue.

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -440,3 +446,67 @@ class ImageSourceCatalogModel(_RomanDataModel, _ParquetMixin):
...
+        l1_shift = models.Shift(-l1_border) & models.Shift(-l1_border)
+        l1_wcs.insert_transform("detector", l1_shift, after=True)
+        bb = wfi_wcs["wcs_l2"].bounding_box
+        if bb is not None:
+            l1_wcs.bounding_box = ((bb[0][0], bb[0][1] + 2 * l1_border), (bb[1][0], bb[1][1] + 2 * l1_border))
+        wfi_wcs.wcs_l1 = l1_wcs
+
+        # Get alignment results, if available
+        if hasattr(model.meta, "wcs_fit_results"):
+            wfi_wcs.meta.wcs_fit_results = copy.deepcopy(model.meta.__getitem__("wcs_fit_results"))
```
braingram, line 509: I just saw that my previous "suggestion" had no suggested change 🤦 
```suggestion
            wfi_wcs.meta.wcs_fit_results = copy.deepcopy(model.meta["wcs_fit_results"])
```

braingram (approved):
Thanks for making the updates to this PR.

I'm marking this as approved since the final suggestion just replaces `__getitem__` with `[]`.

Please update the rad pin before merging.

I also "unresolved" one comment about the GWCS comparison bug:
https://github.com/spacetelescope/roman_datamodels/pull/477/files#r2026929006
Is there a ticket for this?

stscieisenhamer:
About the gwcs comparison bug: no, no issue for that. not sure if that is a gwcs issue or rdm issue, or simply what the state-of-affairs really is.

And about using `[]` in the deepcopy. I tried that initially and did not work; it failed somewhere else but was equally as unhappy.

schlafly (approved):
Good to go pending rad PR merge.

tests/test_models.py
```diff
@@ -1197,3 +1198,76 @@ def test_apcorr_none_array():
...
+    border = 4.0  # Default extra border for L1
+    model_corner = model.meta.wcs.pixel_to_world(0.0, 0.0)
+    wfi_wcs_corner = wfi_wcs.wcs_l1.pixel_to_world(border, border)  # Extra border due to being L1
+    assert model_corner.separation(wfi_wcs_corner).value <= 1e-7
+
+    model_bb = model.meta.wcs.bounding_box
+    wfi_wcs_bb = wfi_wcs.wcs_l1.bounding_box
+    assert model_bb[0][1] + 2 * border == wfi_wcs_bb[0][1]
+    assert model_bb[1][1] + 2 * border == wfi_wcs_bb[1][1]
+
```
PaulHuwe, line 1259: maybe add an `assert wfi_wcs.validate() is None`?

PaulHuwe (approved):
Approved with one possible suggestion

src/roman_datamodels/datamodels/_datamodels.py
stscieisenhamer: Hmmm, cannot replicate though it was the first thing tried. going with it.

stscieisenhamer:
Issue rcal-1052 concerning the gwcs testing issue mentioned above has been created.

PR #488, opened 2025-04-10
Add astropy table metadata to parquet output

schlafly:
This PR adds a new field to the catalog parquet metadata output, table_meta_yaml, to better support astropy's Table.read(...) method.  astropy looks for information in this table to fill out things like table column descriptions and units.

With this PR, astropy tables look like this
```
label     xcentroid          ycentroid         ra_centroid        dec_centroid       aper_bkg_flux       aper_bkg_flux_err   ...       x_psf             x_psf_err             y_psf              y_psf_err            flux_psf          flux_psf_err   
                                                   deg                deg                 uJy                   uJy          ...                                                                                         uJy                 uJy        
int32      float64            float64            float64            float64             float64               float64        ...      float64             float64             float64              float64             float64             float64      
----- ------------------ ------------------ ------------------ ------------------ -------------------- --------------------- ... ------------------ -------------------- ------------------ --------------------- ------------------ -------------------
    1 3107.5641793848517  42.14478665707154 30.102411953272664 44.987236036377105 0.007506220601499081 0.0007226650379779104 ... 3107.5962567666916 0.008446006321761429  42.25312607943128   0.00935998330896384 29.550033445940997  0.3898450679056146
    2   3273.90443956451  46.85926543664257 30.108945192385907   44.9873609690276 0.011104788165539503 0.0009909254350552823 ... 3273.9429928233367 0.009654710513506604 46.991383087661205  0.005123868593818315 47.804684998804674  0.5102734075360951
    3 1027.5074820985774  75.28726347260708  30.02071818407146  44.98820055122564 0.009080739691853523 0.0008211355612551687 ... 1027.6290492829633 0.009623612158641267  75.41621165152297   0.01159528861751296 30.237856484401057   0.401374916753545
```
including the units, while without this they do not have the units.  One can also do things like
```
>>> cat['xcentroid'].description
'X pixel value of the source centroid (0 indexed)'
```

On the other hand, this is a frustrating thing to do because the change adds a new long yaml string to the metadata that recapitulates metadata that is already present in the file, so it's redundant.  But I expect most of our users to get access to these tables via Table.read(...) in astropy and so I think it makes sense overall to make that use case nice.

Thoughts?

### DIFF
--- a/changes/488.feature.rst
+++ b/changes/488.feature.rst
@@ -1,0 +1,1 @@
+Added astropy table metadata to parquet catalog files.
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -7,8 +7,8 @@ This module provides all the specific datamodels used by the Roman pipeline.
 """
 
 import asdf
-import numpy as np
 import astropy.table.meta
+import numpy as np
 from astropy.table import QTable
 
 from roman_datamodels import stnode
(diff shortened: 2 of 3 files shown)


schlafly:
@braingram , I added some round-trip tests.  If you could approve this, I'll merge, thanks!

braingram (approved):
LGTM. We should regenerate a catalog with this new format and send it to the archive. I have this on my list of things to do. I'm hoping to have that file ready today.

PR #499, opened 2025-04-22
backport: fix failing docs (#498) to 0.24.x release branch.

braingram:
Once this is merged I think we need a 0.24.1 release to fix the docs.


schlafly:
Presumably we missed this because with the towncrier stuff we don't actually build the full changelog until ~release time?  Do we need tooling changes to identify these cases before release?

braingram:
I opened https://github.com/spacetelescope/roman_datamodels/issues/500 to track sorting this out.

PR #524, opened 2025-06-02
add __dir__ to allow tab completion

braingram:
This PR implements `__dir__` for data models and nodes. This allows repls to tab complete attributes.

With this PR:
```python
m.me<tab>  # produces m.meta
m.meta.instu<tab>  # produces m.meta.instrument
```

Although this works in the python repl it will not work in ipython (and presumably jupyter) due to design decisions by ipython to:
- explicitly allow-list tab-completion for some dynamic objects (like pandas dataframes and numpy array) and treat other dynamic objects as "unsafe"
- use jedi for completion (which also considering using `__getattr__` unsafe)

Adding the following to your ipython config will allow completion in ipython:
```
c.IPCompleter.evaluation = "unsafe"
c.IPCompleter.use_jedi = False
```

## Tasks

</details


schlafly:
Looks good to me!

Hijacking this PR thread---my personal annoyance along these lines is that I often find myself wanting to get the names of the keys in a tree, and I expect to do something like asdf.open('file.asdf').keys() to do that.  That usually gives me a KeysView(...) of the entire metadata, rather than just the keys, so I end up typing [k for k in asdf.open(...).keys()] to get the actual list of keys.  Sorry, this isn't actually related to this PR, it just reminded me of it.

braingram:
> Looks good to me!
> 
> Hijacking this PR thread---my personal annoyance along these lines is that I often find myself wanting to get the names of the keys in a tree, and I expect to do something like asdf.open('file.asdf').keys() to do that. That usually gives me a KeysView(...) of the entire metadata, rather than just the keys, so I end up typing [k for k in asdf.open(...).keys()] to get the actual list of keys. Sorry, this isn't actually related to this PR, it just reminded me of it.

I share your pain. I often do `list(asdf.open(...).keys())`. Let me try to sort out where it fits to make an issue for that.

PaulHuwe:
Looks great!

docs/roman_datamodels/datamodels/using_datamodels.rst
```diff
@@ -3,6 +3,14 @@
...
 ======================
 
+.. note::
+   IPython and jupyter won't automatically tab-complete the
+   dynamic attributes of datamodels. To enable tab-completion
+   add the following to your `ipython configuration <https://ipython.readthedocs.io/en/stable/config/intro.html>`_::
+
+       c.IPCompleter.evaluation = "unsafe"
+       c.IPCompleter.use_jedi = False
+
```
WilliamJamieson, line 13: This should not be a blanket recommendation. It will break the `jedi` language server which is much more fully featured than the basic one that the repl uses. For example turning this off will mean that you loose some of the static code analysis on object that exist prior to instantiation and it will stop iPython/Jupyter from doing any static type analysis which may surprise users.

PR #538, opened 2025-07-10
Minor bugfixes concerning attributes

### LINKED ISSUE #418: Private metadata attributes are silently lost
```python
>> from roman_datamodels import maker_utils as mk, datamodels as dm
>> m = mk.mk_datamodel(dm.ImageModel)
>> m.meta._foo = 1  # no error
>> m.meta._foo
AttributeError: No attribute _foo
```

I think this is in part due to `_foo` getting set on an ephemeral `DNode` instance.
```python
>> id(m.meta)
4814979664
>> id(m.meta)
5611666880
```

Which leads to this fun line of code:
```python
>> m.meta is m.meta
False
```

Saving a reference to and re-using the `meta` retains `_foo`:
```python
>> mm = m.meta
>> mm._foo = 1
>> mm._foo
1
```

I'm not sure if this is a bug but I was surprised by it.

### LINKED ISSUE #479: Nodes do not implement `__delattr__`
```python
import roman_datamodels.datamodels as rdm, roman_datamodels.maker_utils as mk
m = mk.mk_datamodel(rdm.ImageModel)
del m.meta.ref_file.dark
```

Produces:
```
AttributeError: 'RefFile' object has no attribute 'dark'
```

Expected that "dark" will be deleted from "ref_file". A workaround is to use `__delitem__` (which is implemented):
```
del m.meta.ref_file["dark"]
```

WilliamJamieson:
Closes #418 and closes #479

This PR fixes two small issues with attributes on nodes.

1. Hidden attributes should not be allowed outside of the internal rdm code.
2. `__delattr__` was not defined on nodes and datamodels.

## Tasks

</details

### DIFF
--- a/changes/538.bugfix.rst
+++ b/changes/538.bugfix.rst
@@ -1,3 +1,3 @@
 Minor bugfixes for node and datamodels attributes including, raising explicit errors
-when attemppting to set private variables outside of RDM internal code and adding a
+when attempting to set private variables outside of RDM internal code and adding a
 ``__delattr__`` method to allow the deletion of attributes for both nodes and datamodels.
--- a/changes/538.bugfix.rst
+++ b/changes/538.bugfix.rst
@@ -1,0 +1,3 @@
+Minor bugfixes for node and datamodels attributes including, raising explicit errors
+when attemppting to set private variables outside of RDM internal code and adding a
+``__delattr__`` method to allow the deletion of attributes for both nodes and datamodels.
--- a/tests/test_stnode.py
+++ b/tests/test_stnode.py
@@ -99,16 +99,11 @@ def test_serialization(node_class, tmp_path):
         assert_node_equal(af["node"], node)
 
 
-@pytest.mark.parametrize("node_class", stnode.NODE_CLASSES)
-@pytest.mark.filterwarnings("ignore:Input shape must be 1D")
-@pytest.mark.filterwarnings("ignore:This function assumes shape is 2D")
-@pytest.mark.filterwarnings("ignore:Input shape must be 4D")
-@pytest.mark.filterwarnings("ignore:Input shape must be 5D")
+@pytest.mark.parametrize("node_class", [cls for cls in stnode.NODE_CLASSES if issubclass(cls, stnode.TaggedObjectNode)])
 def test_no_hidden(node_class):
-    if issubclass(node_class, stnode.TaggedObjectNode):
-        node = maker_utils.mk_node(node_class, shape=(8, 8, 8))
-        with pytest.raises(AttributeError, match=r"Cannot set private attribute.*"):
-            node._foo = "bar"  # Add a hidden attribute
+    node = node_class.create_fake_data()
+    with pytest.raises(AttributeError, match=r"Cannot set private attribute.*"):
+        node._foo = "bar"  # Add a hidden attribute
 
 
 def test_info(capsys):
--- a/tests/test_stnode.py
+++ b/tests/test_stnode.py
@@ -99,6 +99,18 @@ def test_serialization(node_class, tmp_path):
         assert_node_equal(af["node"], node)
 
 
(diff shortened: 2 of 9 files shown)


changes/538.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Minor bugfixes for node and datamodels attributes including, raising explicit errors
+when attemppting to set private variables outside of RDM internal code and adding a
```
PaulHuwe, line 2:
```suggestion
when attempting to set private variables outside of RDM internal code and adding a
```

WilliamJamieson:
Regression tests pass with the spacetelescope/romancal#1852 changes: https://github.com/spacetelescope/RegressionTests/actions/runs/16322043303

tests/test_stnode.py
```diff
@@ -113,6 +113,18 @@ def test_serialization(node_class, tmp_path):
...
 
 
+@pytest.mark.parametrize("node_class", stnode.NODE_CLASSES)
+@pytest.mark.filterwarnings("ignore:Input shape must be 1D")
+@pytest.mark.filterwarnings("ignore:This function assumes shape is 2D")
+@pytest.mark.filterwarnings("ignore:Input shape must be 4D")
+@pytest.mark.filterwarnings("ignore:Input shape must be 5D")
+def test_no_hidden(node_class):
+    if issubclass(node_class, stnode.TaggedObjectNode):
+        node = maker_utils.mk_node(node_class, shape=(8, 8, 8))
```
braingram, line 123:
```suggestion
        node = node_class.create_fake_data()
```

tests/test_stnode.py
```diff
@@ -113,6 +113,18 @@ def test_serialization(node_class, tmp_path):
         assert_node_equal(af["node"], node)
 
 
+@pytest.mark.parametrize("node_class", stnode.NODE_CLASSES)
+@pytest.mark.filterwarnings("ignore:Input shape must be 1D")
+@pytest.mark.filterwarnings("ignore:This function assumes shape is 2D")
+@pytest.mark.filterwarnings("ignore:Input shape must be 4D")
+@pytest.mark.filterwarnings("ignore:Input shape must be 5D")
```
braingram, line 120:
```suggestion
@pytest.mark.parametrize("node_class", stnode.NODE_CLASSES)
```

tests/test_stnode.py
```diff
@@ -113,6 +113,18 @@ def test_serialization(node_class, tmp_path):
...
+def test_no_hidden(node_class):
+    if issubclass(node_class, stnode.TaggedObjectNode):
```
braingram, line 122:
```suggestion
    if issubclass(node_class, (stnode.TaggedObjectNode | stnode.TaggedListNode)):
```
Would you move this filtering into the parameterization?

braingram (approved):
Thanks! LGTM pending approval of the romancal PR.

A few suggestions related to the added `test_no_hidden` unit test.

tests/test_stnode.py
WilliamJamieson: This does not apply to list nodes.

braingram: Would you remove the closing verbiage for https://github.com/spacetelescope/roman_datamodels/issues/418 and update it to reflect that the bug still applies for list nodes?

WilliamJamieson: #418 does not indicate any application to list nodes nor would one expect serialization of arbitrary attributes attached to a list to be serialize.

I will add this but I don't think it is necessary

braingram: Thanks. That issue wasn't specific to any node type. It also doesn't have anything to do with serialization. Here's an example with this PR where a private attribute assignment is lost:
```
m.meta.exposure.read_pattern._foo = 1
assert m.meta.exposure.read_pattern._foo == 1
```
Fails with:
```
AttributeError: 'LNode' object has no attribute '_foo'
```
If we're going to prevent private attributes on `DNode` it makes sense to me to enforce the same behavior on `LNode`.

PR #547, opened 2025-08-14
don't put dark.data in dark reference files anymore; it's not required

schlafly:
This removes stops testing for the presence of dark.data and stops populating dark.data in the maker utils, since we no longer want to require it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
   "numpy >=1.24",
   "astropy >=6.0.0",
   # "rad >=0.25.0",
-  "rad @ git+https://github.com/spacetelescope/rad.git",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -17,7 +17,7 @@ dependencies = [
   "numpy >=1.24",
   "astropy >=6.0.0",
   # "rad >=0.25.0",
-  "rad @ git+https://github.com/spacetelescope/rad.git",
+  "rad @ git+https://github.com/PaulHuwe/rad.git@RAD-215_OptionalDarkData",
```
PaulHuwe, line 20:
```suggestion
  "rad @ git+https://github.com/spacetelescope/rad.git",
```

PaulHuwe (approved):

Issue #546, opened 2025-08-12
API for datamodel version migration


braingram:
There are at least a few cases when we would benefit from some migration tooling for converting a file using "old" tags to one that uses all the newest tags (what romancal expects).

A fully general solution is not possible (as new schemas might require data that does not exist in the old file) but sometimes the old and new schemas are 100% compatible and all that is required is "retagging" of a datamodel with the newest tag. We do not have a public API for this trivial case.

The "retagging" could also be interactively used along with calls to `DataModel.validate` to check what migrations are needed for non-trivial cases (let's say a metadata attribute was renamed).

Finally, it's also worth considering (but possibly as a separate issue) if the above API (or other API) could be used after we drop support for some pre-flight versions. Also is this at all useful for TVAC/FPS data?

PR #550, opened 2025-08-25
Add `DataModel.create_from_model` for data migration

braingram:
Resolves https://jira.stsci.edu/browse/RCAL-1136

Closes https://github.com/spacetelescope/roman_datamodels/issues/546

Adds `DataModel.create_from_model` which takes a model (or dictionary) as an argument and constructs a new model (with the newest/default tags) using data from the model. This can be used to:
- update a old file to use the newest tags (although there is no guarantee the old data will be valid with the new schemas)
- construct a model from a provided dictionary. The new method will tag corresponding structures but unlike `create_minimal` and `create_fake_data` will not restrict conversions to required data.
- convert from one model type to another. Note that this does not replace `RampModel.from_science_raw` and `ScienceRawModel.from_tvac_raw` as both of those require additional data changes. A follow-up PR could update these specialized methods to use `create_from_model` (and drop `_node_update`).

Regtests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/17215915665

## Tasks

</details

### DIFF
--- a/src/roman_datamodels/stnode/_schema.py
+++ b/src/roman_datamodels/stnode/_schema.py
@@ -505,14 +505,11 @@ class NodeBuilder(Builder):
     """
     Builder subclass that includes all provided data.
 
-    When constructing objects, values will be determined by (in this order):
-        TODO
-
-    TODO more information
+    When constructing objects, values will be copied from the
+    provided defaults and converted to tags defined in the schema.
     """
 
     def _copy_default(self, default):
-        # TODO is this always needed?
         return copy.deepcopy(default)
 
     def from_unknown(self, schema, defaults):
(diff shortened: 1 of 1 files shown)


schlafly:
Trying to think about when we want to use this... you note cases like from_tvac_raw and from_science_raw, which I agree with.  I feel like we've seen cases along those lines where we want the pipeline to proceed but it fails because of tag mismatches.  You've added code to address that kind of case but I don't remember if there were further cases where we would want this.

I'm trying to decide where we should be using this today.  One thing I could imagine would be for there to be an additional argument to strun or an environment variable that would would ~automatically do something like this, replacing datamodels.open(...) with DataModel.create_from_model(datamodels.open(...)) so that the pipeline always gets the "new" version of a file.  I could also see considering that harmful.

schlafly:
Okay, good.  And yes, Chris Heller reported a successful pipeline installation and initial tests, so I'll probably let @PaulHuwe decide but I think we can go ahead and start making serious changes again.

PR #555, opened 2025-08-30
Follow on for removal of tagged scalars from RAD

WilliamJamieson:
This PR includes the changes for `roman_datamodels` that are necessary to support spacetelescope/rad#688. The vast majority of the changes are simply the result of needing to control when scalars need to be wrapped and when they don't. For legacy reasons with things like TVAC and FPS we still need to support auto wrapping of the `TaggedScalar` values.

Note that since we no longer have a specific tagged object for a special class to be involved. The `create_minimal` mixins used for a few of the tagged scalars to assign values no longer works. Those values appear to be pretty much the same always so I moved the setting of them to the datamodel initializer.  I am open to suggestions there.

## Tasks

</details

### DIFF
--- a/src/roman_datamodels/datamodels/_datamodels.py
+++ b/src/roman_datamodels/datamodels/_datamodels.py
@@ -148,6 +148,13 @@ class _RomanDataModel(_DataModel):
 
         if init is not None:
             self.meta.model_type = self.__class__.__name__
+            self.meta.calibration_software_name = "RomanCAL"
+
+            if "file_date" not in self.meta:
+                self.meta.file_date = Time.now()
+
+            if "origin" not in self.meta:
+                self.meta.origin = "STSCI/SOC"
 
 
 class MosaicModel(_RomanDataModel):
--- a/src/roman_datamodels/stnode/_node.py
+++ b/src/roman_datamodels/stnode/_node.py
@@ -38,22 +38,20 @@ class DNode(MutableMapping):
         self._parent = parent
         self._name = name
 
-    @staticmethod
-    def _convert_to_scalar(key, value, ref=None):
-        """Find and wrap scalars in the appropriate class, if its a tagged one."""
-        from ._tagged import TaggedScalarNode
+    def _wrap_value(self, key, value):
+        # Return objects as node classes, if applicable
+        if isinstance(value, dict | AsdfDictNode):
+            return DNode(value, parent=self, name=key)
 
-        if isinstance(ref, TaggedScalarNode):
-            # we want the exact class (not possible subclasses)
-            if type(value) == type(ref):  # noqa: E721
-                return value
-            return type(ref)(value)
+        elif isinstance(value, list | AsdfListNode):
+            return LNode(value)
 
-        if isinstance(value, TaggedScalarNode):
+        else:
             return value
 
-        if key in SCALAR_NODE_CLASSES_BY_KEY:
-            value = SCALAR_NODE_CLASSES_BY_KEY[key](value)
+    @staticmethod
+    def _convert_to_scalar(key, value, ref=None):
+        """For now this is a simple pass through so a mixin can override it."""
 
         return value
 
@@ -73,14 +71,7 @@ class DNode(MutableMapping):
             value = self._convert_to_scalar(key, self._data[key])
 
             # Return objects as node classes, if applicable
-            if isinstance(value, dict | AsdfDictNode):
-                return DNode(value, parent=self, name=key)
-
-            elif isinstance(value, list | AsdfListNode):
-                return LNode(value)
-
-            else:
-                return value
+            return self._wrap_value(key, value)
 
         # Raise the correct error for the attribute not being found
         raise AttributeError(f"No such attribute ({key}) found in node")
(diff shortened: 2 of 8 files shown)


src/roman_datamodels/stnode/_node.py
```diff
@@ -38,22 +38,20 @@ def __init__(self, node=None, parent=None, name=None):
...
+        else:
             return value
 
-        if key in SCALAR_NODE_CLASSES_BY_KEY:
-            value = SCALAR_NODE_CLASSES_BY_KEY[key](value)
+    @staticmethod
+    def _convert_to_scalar(key, value, ref=None):
+        """For now this is a simple pass through so a mixin can override it."""
 
         return value
```
WilliamJamieson, line 56: We may want to indicate off the `_read_tag` so that old files still function the same.

WilliamJamieson:
Note this PR is based on top of #538 so that the `_read_tag` attribute is always available. This allows for more precise control of when wrapping occurs, this will still allow wrapping for older tag versions but not for the newest versions.

PaulHuwe (approved):
LGTM modulo the question

src/roman_datamodels/stnode/_node.py
```diff
@@ -4,56 +4,78 @@
...
 
-from ._registry import SCALAR_NODE_CLASSES_BY_KEY
+from ._registry import INTERNAL_WRAP_LIMITS, SCALAR_NODE_CLASSES_BY_KEY
 
-__all__ = ["DNode", "LNode"]
+__all__ = ["DNode", "LNode", "TaggedScalarDNode"]
+
+
+def _convert_to_scalar(key, value, ref=None):
+    """Wrap scalars in into a tagged scalar object"""
```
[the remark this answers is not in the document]
WilliamJamieson: No this is wrapping something into a tagged scalar. It is called for legacy support

src/roman_datamodels/stnode/_node.py
```diff
@@ -4,56 +4,78 @@
...
 
         # Set the metadata tracked by the node
         self._parent = parent
         self._name = name
+        self._read_tag = None
 
-    @staticmethod
-    def _convert_to_scalar(key, value, ref=None):
+    def _convert_to_scalar(self, key, value, ref=None):
         """Find and wrap scalars in the appropriate class, if its a tagged one."""
```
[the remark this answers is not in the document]
WilliamJamieson: Is literally, a modification of an existing method.

src/roman_datamodels/datamodels/_utils.py
```diff
@@ -302,3 +319,64 @@ def rdm_open(init, memmap=False, **kwargs):
...
+    )
+    output.validate()
+
+    def _callback(node):
+        if isinstance(node, TaggedScalarNode):
+            return node.unwrap()
+        return node
+
+    new = type(output).create_from_model(asdf.treeutil.walk_and_modify(dict(output._instance), _callback))
+    new._instance._read_tag = _INTERNAL_TAG_REMOVAL_MAP[tag]
```
braingram, line 380:
```suggestion
    output.validate()

    def _callback(node):
        if isinstance(node, TaggedScalarNode):
            return node.unwrap()
        return node

    new = type(output).create_from_model(asdf.treeutil.walk_and_modify(dict(output._instance), _callback))
    new._instance._read_tag = _INTERNAL_TAG_REMOVAL_MAP[tag]
```
Would you split the `auto_upgrade` feature into a new PR? There are a few issues here:
- should the default be `True`?
- should the upgraded model be validated?
- how can this be done efficiently? (the current code reads the entire input model into memory, creates a copy via walk_and_modify, creates a second copy via create_from_model)

I'd also like to see what romancal tests fail without this auto_upgrade. Can these all be addressed by regenerating regtest files?

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -156,6 +156,13 @@ def __init__(self, init=None, **kwargs):
 
         if init is not None:
             self.meta.model_type = self.__class__.__name__
+            self.meta.calibration_software_name = "RomanCAL"
+
+            if "file_date" not in self.meta:
+                self.meta.file_date = Time.now()
+
+            if "origin" not in self.meta:
+                self.meta.origin = "STSCI/SOC"
```
braingram, line 165:
```suggestion
```
Let's handle this in the mixin classes. The code here populates/overwrites these values for all models. With this branch if I save a model where I've set `calibration_software_name` to a different value it incorrectly reads back in as `RomanCal`.

braingram (changes requested):

src/roman_datamodels/datamodels/_datamodels.py
WilliamJamieson: This will require a mixin for every datamodel that isn't a reference file.

WilliamJamieson: You understand that code better than I do. So feel free to push a solution to my branch.

WilliamJamieson: I think I have a solution by passing "default set of defaults" for the datamodel products.

src/roman_datamodels/stnode/_node.py
```diff
@@ -10,10 +10,30 @@
...
+__all__ = ["DNode", "LNode", "TaggedScalarDNode"]
+
+
+def _convert_to_scalar(key, value, ref=None):
+    """Wrap scalars in into a tagged scalar object"""
+    from ._tagged import TaggedScalarNode
+
+    if isinstance(ref, TaggedScalarNode):
+        # we want the exact class (not possible subclasses)
+        if type(value) == type(ref):  # noqa: E721
```
braingram, line 26:
```suggestion
        if type(value) is type(ref):
```
Does `is` work here and allow the noqa removal?

WilliamJamieson: I think so, good catch! I was just copying existing code around and into a new function.

PR #566, opened 2025-09-18
Enable MyPy and fix the small number of errors it finds.

WilliamJamieson:
As suggested by this comment, https://github.com/spacetelescope/roman_datamodels/pull/565#discussion_r2360750952, this PR enables mypy testing and fixes all the type errors it highlighted.

## Tasks
- [ ] Update or add relevant `roman_datamodels` tests.
- [ ] Update relevant docstrings and / or `docs/` page.
- [ ] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [ ] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [ ] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

<details><summary>News fragment change types:</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details

PR #568, opened 2025-09-19
Fix some minor things I noticed while doing the mypy PR

WilliamJamieson:
This PR fixes a couple of things that I noticed while doing the mypy PR #566. These didn't fit with that theme nor do they fit anywhere else.

## Tasks

</details

### DIFF
--- a/changes/568.feature.rst
+++ b/changes/568.feature.rst
@@ -1,0 +1,6 @@
+Minor improvements to stnode and the datamodels to make them easier for
+static analysis tools like mypy to understand.
+
+Slight API change to stnode to hide the ``builder`` parameter in the ``create_*``
+methods. This parameter was only intended for the internal implementation of
+these methods, and not for public use. The other changes uncovered this issue.


src/roman_datamodels/stnode/_tagged.py
```diff
@@ -45,7 +43,108 @@ def name_from_tag_uri(tag_uri):
     return tag_uri_split
 
 
-class TaggedObjectNode(DNode):
+class _TaggedNodeMixin(ABC):
```
WilliamJamieson, line 46: All the tagged objects share all of these and had copies of them so I pulled them out into this.

src/roman_datamodels/stnode/_node.py
```diff
@@ -29,9 +25,6 @@ class DNode(MutableMapping):
 
     __slots__ = ("_data", "_name", "_parent", "_read_tag")
 
-    _pattern: ClassVar[str]
-    _latest_manifest: ClassVar[str]
```
WilliamJamieson, line 33: These were only useful to the tagged nodes so I put them there, note that `_read_tag` will be used here after we do the tagged scalars, so I didn't move it.

src/roman_datamodels/stnode/_mixins.py
```diff
@@ -13,16 +13,20 @@
...
 if TYPE_CHECKING:
     from typing import ClassVar, TypeAlias
 
     from astropy.time import Time
 
     from ._tagged import TaggedObjectNode, TaggedScalarNode
 
+    class _TimeNode(Time, TaggedScalarNode):
+        pass
+
```
WilliamJamieson, line 26: Hacky way to do this but you can't double inherit from the same object nor is mypy happy with a tuple unpack.

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -131,30 +129,7 @@ def to_parquet(self, filepath):
...
-        if cls.__name__.startswith("_"):
-            return
-
-        if cls.__name__ in __all__:
-            raise ValueError(f"Duplicate model type {cls.__name__}")
-
-        __all__.append(cls.__name__)
-
-
-class _RomanDataModel(_DataModel):
```
WilliamJamieson, line 157: I noticed that all this along with the __all__ was blocking my IDE from back tracing to the datamodel objects here. This isn't necessary because the  _-prefix will block against those being part of __all__.

braingram (commented):
Overall LGTM but docs are failing would you take a look at what's causing the failure? Also would you run the regtests with this branch when it's updated to fix the docs?

src/roman_datamodels/datamodels/_datamodels.py
```diff
@@ -27,8 +27,6 @@
 if TYPE_CHECKING:
     from typing import Any
 
-__all__ = []
```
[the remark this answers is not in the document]
braingram: Thanks. Would you add a comment here describing why this file does not have an `__all__`?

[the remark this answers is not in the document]
WilliamJamieson: https://github.com/WilliamJamieson/roman_datamodels/blob/9f7af7367480a518b98421ef4cda73ef5a585d60/src/roman_datamodels/datamodels/_datamodels.py#L15 and https://github.com/WilliamJamieson/roman_datamodels/blob/9f7af7367480a518b98421ef4cda73ef5a585d60/src/roman_datamodels/datamodels/_datamodels.py#L20 were changed to be namespace imports to avoid having the `sphinx-automodapi` trying to add those to the API docs with this change.

WilliamJamieson:
Everything passes https://github.com/spacetelescope/RegressionTests/actions/runs/17922075901/job/50959427064

src/roman_datamodels/stnode/_schema.py
```diff
@@ -14,7 +14,7 @@
     SCHEMA_URIS_BY_TAG,
 )
 
-__all__ = []
+__all__ = ["Builder"]
```
braingram, line 17: Why is this being added to the public API?

WilliamJamieson: Why not add a `Builder` protocol? 

```python
from typing import Protocol

class Builder(Protocol):

    def build(self, schema: Mapping[str, Any], defaults: Any | None = None):
         """Method to build a map based on a ``schema`` with values overridden by the passed``defaults``""""

```

WilliamJamieson: I did not add any method called `create_from_node`. That was added by you in #550 so it was already public API. As far as I am aware I didn't actual "add" or even change public API with this PR as the Protocol is simply a description of what was expected to be passed to a public function already.

src/roman_datamodels/stnode/_schema.py
```diff
@@ -187,7 +191,32 @@ class SchemaType(enum.Enum):
...
+            The defaults to use when building.
+
+        Returns
+        -------
+        Mapping[str, Any] or Sequence[Any]
+            The built data for a node based on the schema using the defaults where possible.
+        """
+
+
+class Builder(Build):
```
WilliamJamieson, line 219: Not that inheritance from a `Protocol` is the same as inheriting from `abc.Abstract`; however, one does not have to inherit from it if one does not have to. The `@runtime_checkable` just allows `isinstance` to check if an instance of an object satisfies the protocol. In fact all the `collections.abc` are actually just `Protocol` objects with default implementations for many of the methods.

src/roman_datamodels/stnode/_schema.py
braingram: Thanks. Yeah that was my bad on `create_from_node` I lost track of that change with the addition of `_TaggedNodeMixin` in this PR.

I do not think `Builder` should be public. If that's a sticking point for this PR we can put this on hold until the arguments to the `create_*` functions can be renamed to `_builder`.

src/roman_datamodels/stnode/_schema.py
```diff
@@ -187,7 +191,32 @@ class SchemaType(enum.Enum):
 _NUMERIC_KEYWORDS = {"multipleOf", "maximum", "exclusiveMaximum", "minimum"}
 
 
-class Builder:
+@runtime_checkable
+class Build(Protocol):
```
braingram, line 195: I do not think this should be public.

braingram (dismissed):

WilliamJamieson:
@braingram commit 1a47c73b516972087289783c9a459d2bd0fc61d2 of this PR is simply what I linked to in https://github.com/spacetelescope/roman_datamodels/pull/569#discussion_r2373357243. I am happy for you to us this for #569, or some other similar change or we can simply include it as part of this PR. It just made sense to rebase these changes on those as otherwise the merge conflicts will get annoying to deal with.

src/roman_datamodels/stnode/_tagged.py
```diff
@@ -78,12 +85,82 @@ def create_minimal(cls, defaults=None, builder=None, *, tag: str | None = None):
...
+        shape: tuple[int, ...] | None = None,
+        builder: Builder | None = None,
+        *,
+        tag: str | None = None,
+    ) -> Self:
+        return cls._create_minimal(defaults, builder or FakeDataBuilder(shape), tag=tag)
+
+    @classmethod
+    def create_fake_data(
+        cls, defaults: Mapping[str, Any] | None = None, shape: tuple[int, ...] | None = None, *, tag: str | None = None
```
WilliamJamieson, line 120:
```suggestion
        cls, *, defaults: Mapping[str, Any] | None = None, shape: tuple[int, ...] | None = None, tag: str | None = None
```

Ideally, I would prefer to make these keyword only inputs as things like this function really don't have a good defined of inputs you would want to pass. It seems just as prudent to pass `shape` only as it is `defaults` only.  In general, I am in favor of making function parameters with default values keyword only whenever possible.

WilliamJamieson:
> > @braingram commit [1a47c73](https://github.com/spacetelescope/roman_datamodels/commit/1a47c73b516972087289783c9a459d2bd0fc61d2) of this PR is simply what I linked to in [#569 (comment)](https://github.com/spacetelescope/roman_datamodels/pull/569#discussion_r2373357243). I am happy for you to us this for #569, or some other similar change or we can simply include it as part of this PR. It just made sense to rebase these changes on those as otherwise the merge conflicts will get annoying to deal with.
> 
> Thanks. I haven't had time to address the requested changes in #569. I have no objection if you updated this PR with a changelog describing the changes in [1a47c73](https://github.com/spacetelescope/roman_datamodels/commit/1a47c73b516972087289783c9a459d2bd0fc61d2) and I'll close #569.

Done.

src/roman_datamodels/stnode/_tagged.py
```diff
@@ -45,30 +42,40 @@ def name_from_tag_uri(tag_uri):
...
-        Register any subclasses of this class in the OBJECT_NODE_CLASSES_BY_PATTERN
-        registry.
-        """
-        super().__init_subclass__(**kwargs)
-        if cls.__name__ != "TaggedObjectNode":
-            if cls._pattern in OBJECT_NODE_CLASSES_BY_PATTERN:
-                raise RuntimeError(f"TaggedObjectNode class for tag '{cls._pattern}' has been defined twice")
-            OBJECT_NODE_CLASSES_BY_PATTERN[cls._pattern] = cls
+    def __init__(self, *args, **kwargs):
+        pass
```
braingram, line 73:
```suggestion
    def __init__(self, *args, **kwargs):
        pass
```
Is this needed?

braingram (commented):
Just one question about the empty `__init__`.

src/roman_datamodels/stnode/_tagged.py
WilliamJamieson: Yes, MyPy needs it down on lines 80 and 253, so that it knows the "class initializer" accepts inputs. Since this is immediately overloaded and the mixin is never used on its own, this code is only there to make MyPy happy.

braingram: If there's a solution that involves telling mypy to ignore this I'd prefer that over this empty `__init__`. If not, would you add a comment describing why it needs to be here to appease mypy?

WilliamJamieson: I ported part of the changes for my solution to this PR. It is extremely minor.

PR #574, opened 2025-09-29
Explicitly build `individual_image_meta` tables from schemas

WilliamJamieson:
Closes #559

This PR takes advantage of the changes in spacetelescope/rad#697 to enable RDM to parse the `individual_image_meta` schema so that it can explicitly parse out the information from each image into the tables based on the structure of the schemas rather than the current implicit table construction found in https://github.com/spacetelescope/romancal/blob/d97b36297c715c1cfc0c6c2a0f7d9a50ec03c4d1/romancal/resample/meta_blender.py which is implicitly building these tables.

This PR is here in RDM as it involves direct schema parsing, but this PR does not replace the `meta_blender` in RCAL completely. It only takes care of the table construction as it relates to the layout of the tables defined by RAD.

## Tasks

</details


WilliamJamieson:
> I will aim to take a closer look at this and the linked PRs this week.
> 
> In general, since there is no expectation that users will be blending metadata (calling `MosaicModel.add_image`) I think it makes the most sense to keep the blending code in romancal and not to move it back into roman_datamodels.

I disagree, this code is parsing schemas which is the primary domain of RDM. Moving it into RCAL will require RCAL to have a direct dependency on RAD, which leads to other issues. We can simply make the API private if it so concerns you. The rest of it is hidden. Best practice simply indicates that we should consider making it public API simply because there is cross package usage.

Moreover there is no metadata blending there is simple table assembly, this is effectively just accounting. All the metadata blending is within RCAL only.

PR #575, opened 2025-09-29
Updates to support internal tag removals

WilliamJamieson:
This PR is the follow on RDM changes to support spacetelescope/rad#691. Note that this depends on #574 so that the `individual_image_meta` tables get properly built now that tags are missing in RCAL. For the most part these changes are minor and expected.

## Tasks

</details


tests/test_stnode.py
```diff
@@ -104,14 +107,30 @@ def test_wfi_mode():
...
+        #   will have `read_tag` populated and since it is a pre-internal tag removal schema
+        #   it will run the wrapping code that causes the `product_type` to be autop wrapped
+        #   into the `ProductType` tagged scalar node which means it is not longer a simple string
+        #   like it really should be. `create_fake_data` is not inducing the wrapper because
+        #   it is not setting the `read_tag` attribute, so auto wrapping is not happening.
+        # This is technically a subtle RDM bug for the old internal tagged objects, but this is the
+        #   only case and it would only effect things taking advantage of the old code.
+        request.applymarker(
+            pytest.mark.xfail(reason="mosaic basic has a product_type not derived from the same schema as product_type")
+        )
```
WilliamJamieson, line 132: It should be possible to resolve this bug, but it is effectively only detectable at this level since the `ProductType` object is simply a `str` subclass with some metadata attributes added.

After trying to solve this cleanly, I don't think it is worth the effort or complexity necessary to correct this. A better solution would be to do as the linked RAD issue suggests and align the two `product_type` definitions so that it can be a single direct reference.

WilliamJamieson:
Regression tests look like their failing: https://github.com/spacetelescope/RegressionTests/actions/runs/18165225603/job/51705798130. However, all the failures are simply expected due to the type changes this will induce.

tests/test_stnode.py
braingram: Would you remove this xfail and update the following line:
https://github.com/spacetelescope/roman_datamodels/blob/69d0823f2875f6ed1a855e35cbf935c5ad0e0d7b/src/roman_datamodels/stnode/_stnode.py#L66
to
```python
        if 'mosaic_basic' not in base and manifest_uri == _LAST_INTERNAL_TAGGED_MANIFEST:
```
It looks like the current code on main is incorrectly wrapping product_type for old `MosaicBasic` instances. From the schemas this should never be wrapped so there is no need to add a wrap limit for that schema/node.

WilliamJamieson: Done

pyproject.toml
```diff
@@ -17,7 +17,8 @@ dependencies = [
   "numpy >=1.24",
   "astropy >=6.0.0",
   # "rad >=0.25.0",
-  "rad @ git+https://github.com/spacetelescope/rad.git",
+  # "rad @ git+https://github.com/spacetelescope/rad.git",
+  "rad @ git+https://github.com/WilliamJamieson/rad.git@remove_internal_tags",
```
braingram, line 21:
```suggestion
  "rad @ git+https://github.com/spacetelescope/rad.git",
```
Reminder to update prior to merging.

tests/test_models.py
```diff
@@ -923,7 +927,9 @@ def test_create_from_model_old_tags():
 
     converted = datamodels.ImageModel.create_from_model(old_model)
     assert converted.tag == new_model_tag
-    assert converted.meta.observation.tag == new_observation_tag
+    # New models should notg have a tagged observation node
```
PaulHuwe, line 930:
```suggestion
    # New models should not have a tagged observation node
```

PaulHuwe (commented):
Looks good with one change.

PaulHuwe (approved):
LGTM

PR #621, opened 2026-01-20
[SCSB-236] use Python 3.13 to build docs with Sphinx 9

zacharyburnett:
Resolves [SCSB-236](https://jira.stsci.edu/browse/SCSB-236)


braingram (approved):
Thanks! A scan of the built docs looks good to me.

Issue #655, opened 2026-04-23
Do we benefit from dynamic node creation?


braingram:
As we no longer use nested node (each datamodel = 1 node) and are considering dropping support for old multi-node datamodels we may also want to consider if the dynamic creation of nodes is helpful. 

At this point any new nodes would only be introduced if we introduce a new datamodel. In that case we currently still have to make paired rad/roman_datamodels PRs since datamodel creation is not dynamic.

It seems straightforward to:
- freeze the current dynamic nodes
- deprecate/drop those we don't want to support in the future

The benefits are:
- significantly simplify roman_datamodels internals
- remove many of the "mixin" classes which complicate datamodel maintenance by spreading behavior across files and objects
- (likely) reduce the overhead of roman_datamodels import (which could be a minor issue for multiprocessing where each process may be subject to the import overhead)
- static code inspection of nodes would be possible

The cons are:
- significant changes to roman_datamodels internals requiring time from all maintainers
- minor increase in code needed to support a new datamodel (as a new node will also need to be created) for simple cases this would be an extra 3 or 4 lines of code

This would have no impact on public API.

braingram:
Another benefit discussed is affording more flexibility for node<->class mapping for dropping support or special handling of tags like in:
https://github.com/spacetelescope/rad/issues/869
https://github.com/spacetelescope/rad/issues/860

PR #658, opened 2026-04-29
Remove the LNode class

WilliamJamieson:
The LNode added significant complexity and was only really used by the legacy support TaggedListNode, which only handled a single legacy object.

This PR  removes LNode as a distinct object and moves the minimal implementation to support TaggedListNode's functionality directly into that object. This is part of an effort to reduce the complexities of the internal stnode subpackage

This will help when we remove dynamic node creation suggested in #655.

## Tasks

</details

### DIFF
--- a/changes/658.removal.rst
+++ b/changes/658.removal.rst
@@ -1,0 +1,1 @@
+Remove the ``LNode`` class and refactor its functionality into the ``TaggedListNode`` class.
--- a/docs/roman_datamodels/datamodels/stnode.rst
+++ b/docs/roman_datamodels/datamodels/stnode.rst
@@ -2,28 +2,36 @@ Stnode Functionality
 ====================
 
 As seen throughout :ref:`using-datamodels`, "node" objects are used to actually
-handle and store the data for a datamodel. Indeed, if one opens an ASDF file
+handle and store the data for a data model. Indeed, if one opens an ASDF file
 with a roman datamodel directly with `asdf.open` instead of
 `roman_datamodels.datamodels.open`, the resulting object stored under the
-``roman`` attribute will be a specialized `~roman_datamodels._stnode.DNode` object rather
-than a `~roman_datamodels.datamodels.DataModel` object. Thus the stnode, "node"
-objects form the data storage and manipulation backbone of the roman datamodels.
+``roman`` attribute will be a specialized `~roman_datamodels._stnode.DNode` object
+rather than a `~roman_datamodels.datamodels.DataModel` object. Thus the stnode,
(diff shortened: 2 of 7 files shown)


WilliamJamieson:
The regression tests pass except for the failures fixed by the above linked RCAL and romanisim PRs:
https://github.com/spacetelescope/RegressionTests/actions/runs/25121815195/job/73624284516

It looks like a new warning in romanisim is blocking testing for RCAL with the updated changes, see https://github.com/spacetelescope/romancal/actions/runs/25124993929/job/73635598639?pr=2288

braingram (changes requested):
I think we should keep `LNode` since it provides a consistent interface for metadata within lists. One example of this is in the multiband segmentation map. On main:
```python
m = rdm.open("roman-pipeline/dev/truth/WFI/image/r00001_p_v01001001001001_270p65x70y49_segm.asdf")
m.meta.image_metas[0].wcsinfo.ra
```
works as expected and gets the ra for the first image that contributed to the data. With this PR the dotted access no longer works after the 0-th indexing and instead a user would need to use:
```
m.meta.image_metas[0]["wcsinfo"]["ra"]
```

schlafly:
I'll defer to you two.  I did spend a little trying to understand some LNode related issue yesterday and was optimistic that this PR would make it go away.  That issue:
```
l1 = open_dataset('/grp/roman/eschlafly/romanimpreprocess/sample_files/sim_L1_J129_4086_12.asdf', update_version=True)
l1.save('blah.asdf')  # works
l1.extrameta = dict(read_pattern=l1.meta.exposure.read_pattern)
l1.save('blah.asdf')  # bugs out trying to serialize an Lone
l1.extrameta = dict(read_pattern=[list(x) for x in l1.meta.exposure.read_pattern])
l1.save('blah.asdf')  # works
```

This is with old L1 files from the HLIS PIT.

braingram:
The lack of node serialization extends beyond `LNode`.
```python
m.extra = {"extra": m.meta.exposure}
m.save("foo.asdf")  # fails since now the tree has an errant DNode
```
Perhaps we should add a `DNode`/`LNode` converter? That would be straightforward to do but would touch some of the internals that I was told shouldn't be touched until we have plans for other issues.

schlafly:
Thanks.  And the right way to get around this at present is to do conversions to list / dict before writing?

braingram:
```python
l1.extrameta = dict(read_pattern=l1.meta.exposure["read_pattern"])
```
or
```python
m.extra = {"extra": m.meta["exposure"]}
```
should work (tested the second one and the file is saveable)

schlafly:
Thanks, I can confirm that works.

PR #659, opened 2026-04-29
Make create_fake_data use provided list defaults (for example read_pattern)

braingram:
While looking at https://github.com/spacetelescope/romancal/pull/2287 I noticed that providing a `read_pattern` to `create_fake_data` did not work as expected. The provided `read_pattern` was ignored and instead was always an empty list.

This PR updates the builder backing `create_fake_data` to use the provided list defaults if the schema-derived fake is empty. This mostly aligns with how object/dict creation works (where the default is used if the schema doesn't define a value). However the list creation doesn't attempt to merge the defaults with any schema-derived fake list (since this seems overly complicated and error-prone).

We have one test:
https://github.com/spacetelescope/roman_datamodels/blob/649c8ebe1789e0ca02af6e1fb18316f84e86a4ad/tests/test_models.py#L611-L618
that was silently passing but was being impacted by the bug fixed in this PR. We might want to update that test to include something like:
```python
assert ramp.meta.exposure.read_pattern == [[1], [2], [3]]
```
but given the conversation about the future of `from_science_raw` we might wait until that is resolved.

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/25123011350

all pass

## Tasks

</details

### DIFF
--- a/changes/659.bugfix.rst
+++ b/changes/659.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bug where list defaults were ignored by ``create_fake_data``.
--- a/src/roman_datamodels/_stnode/_schema.py
+++ b/src/roman_datamodels/_stnode/_schema.py
@@ -485,6 +485,13 @@ class FakeDataBuilder(Builder):
                     obj[k] = copy.deepcopy(v)
         return obj
 
+    def from_array(self, schema, defaults):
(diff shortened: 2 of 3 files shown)


schlafly (approved):
Thanks Brett.

PR #668, opened 2026-05-05
Remove dynamic node creation for each tag in RAD

### LINKED ISSUE #655: Do we benefit from dynamic node creation?

### LINKED ISSUE #657: Improve (or remove) internal "registries"
Follow up from: https://github.com/spacetelescope/roman_datamodels/pull/656

The "registries" inside stnode are difficult to use due to:
- the population of registries occurs in a separate file from where the registries are defined making it difficult to trace and coordinate how a registry is populated
- their mutability has led to previous issues where unintentional changes were had to track down

They can be removed with minor code restructuring as seen in the linked closed PR but will touch code that is likely to be changed while addressing:
- https://github.com/spacetelescope/rad/issues/860
- https://github.com/spacetelescope/rad/issues/869
- https://github.com/spacetelescope/roman_datamodels/issues/655

WilliamJamieson:
This PR removes the dynamic node creation from stnode that was being performed for each tag_uri lineage in RAD. Now the only dynamic creation is for the RAD manifests, which is only done to support having the history recorded correctly in the ASDF file.

Note this is based on #675

closes #655

## Tasks

</details


WilliamJamieson:
We are not planning on making any changes at this time.

PR #669, opened 2026-05-05
Remove `_latest_manifest` and `_default_tag` ClassVar from stnode

WilliamJamieson:
This PR removes the `_latest_manifest` and `_default_tag` class variables from stnode and replaces them with lookups based on the `_pattern` class variable. This is part of an effort to reduce the object complexity in stnode.

Note that this PR depends on #671

## Tasks
- [ ] Update or add relevant `roman_datamodels` tests.
- [ ] Update relevant docstrings and / or `docs/` page.
- [ ] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [ ] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [ ] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

<details><summary>News fragment change types:</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details

PR #664, opened 2026-04-30
No more stnode mixins

WilliamJamieson:
This PR eliminates all the objects surrounding the `._stnode._mixins` and replaces them with a combination of the explicit `TaggedScalarNode` subclasses `TaggedStrNode` and `TaggedTimeNode` in addition to moving the source catalog related code into the data model and builder classes. 

## Tasks
- [ ] Update or add relevant `roman_datamodels` tests.
- [ ] Update relevant docstrings and / or `docs/` page.
- [ ] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [ ] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [ ] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

<details><summary>News fragment change types:</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details

PR #663, opened 2026-04-30
Switch data models to use tag_patterns rather than node_type

WilliamJamieson:
In order to make stnode less dynamic and less registry dependent, the first step is to swap the datamodels from using the _node_type to just using the _tag_pattern and using that to align the types.

As a benefit this eliminates the need to keep a type stubs file for stnode 

## Tasks
- [ ] Update or add relevant `roman_datamodels` tests.
- [ ] Update relevant docstrings and / or `docs/` page.
- [ ] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [ ] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [ ] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

<details><summary>News fragment change types:</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details

PR #660, opened 2026-04-30
Update `tag` handling and clean up `Builder` API

WilliamJamieson:
There has been something about how stnode objects handle the setting and removal of the tags that ASDF (or users) can attach to nodes that has been bothering me for a long time, see https://github.com/spacetelescope/roman_datamodels/blob/649c8ebe1789e0ca02af6e1fb18316f84e86a4ad/src/roman_datamodels/_stnode/_converters.py#L69.

This PR attempts to resolve this by adding a constructor to the nodes that takes care of this issue. This in turn means that all the code of the form
```python
if tag:
    new._read_tag = tag

return new
```

could be rolled together. Note that while fixing this issue I noticed/found some small issues with the `Builder`/`create_*` API that could be improved. This PR resolves both the `_read_tag` and builder API stuff.

## Tasks
- [ ] Update or add relevant `roman_datamodels` tests.
- [ ] Update relevant docstrings and / or `docs/` page.
- [ ] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [ ] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [ ] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

<details><summary>News fragment change types:</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details

PR #673, opened 2026-05-06
Move the default tag system into the datamodel

WilliamJamieson:
If/when we remove dynamically generated node classes, the node classes will no longer be able to determine their own default tags (since it will be generic to all tags). This PR moves the default tag handling from the node class into the data model so that the nodes can be made generic with respect to the actual tag involved.

Note this depends on #660, #663, #664, #669, and #670

## Tasks

</details


WilliamJamieson:
We are not planning on making any changes at this time.

PR #678, opened 2026-05-15
release metadata for version `1.0.0`

zacharyburnett:
blocked by `rad 1.0.0` release

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
   "gwcs >=0.20.0",
   "numpy >=1.25",
   "astropy >=6.0.0",
-  "rad @ git+https://github.com/spacetelescope/rad.git",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -16,7 +16,7 @@ dependencies = [
   "gwcs >=0.20.0",
   "numpy >=1.25",
   "astropy >=6.0.0",
-  "rad @ git+https://github.com/spacetelescope/rad.git",
+  "rad >=1.0,<1.1",
```
braingram, line 19: Pinging @spacetelescope/roman_datamodels-maintainers for awareness.
@zacharyburnett and I discussed this during tag-up. I believe this follows the discussed version scheme where since rad and rdm and essentially paired we want to pin on the **minor** version here. Please chime in if folks think otherwise.

braingram (approved):

PR #686, opened 2026-07-03
Set array_inline_threshold to save small arrays in the YAML tree.

### LINKED ISSUE #681: Set array_inline_threshold to store wcs (and other small) arrays inline instead of in compressed blocks
Do we want to set a value for `array_inline_threshold`?
https://www.asdf-format.org/projects/asdf/en/latest/api/asdf.config.AsdfConfig.html#asdf.config.AsdfConfig.array_inline_threshold

Testing with a level 2 file, setting a threshold of 1k results in 15 instead of 27 blocks in the level 2 file with the 12 arrays used by the WCS stored inline. This has a few benefits:

- reading the wcs does not require decompressing data
- reading the wcs does not require parsing blocks (as data is inlined) (which may make the wcs easier to pass off to other libraries like AST which support YAML parsing but not ASDF blocks)

the file size didn't significantly change:
- old: 137384895 bytes
- new: 137386670 (+1775)

The test level 3 file I used shows a slightly different story:
- old: 140 blocks, 508015657 bytes
- new: 6 blocks, 508003137 (-12520)

This is largely due to the tables used to store the level 2 metadata within the level 3 file (only 3 level 2 file for this level 3 which is likely on the small side).

braingram:
Closes: https://github.com/spacetelescope/roman_datamodels/issues/681

Sets `array_inline_threshold` to trigger asdf to save small arrays (<512 bytes) inline (not in a binary block).

Some test code changes were needed to work around some asdf bugs (that are being addressed). Since these bugs only impact test code it seems reasonable to use the work arounds.

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/28631348130/job/84909023280
all pass

## Tasks

</details

### DIFF
--- a/src/roman_datamodels/_stnode/_schema.py
+++ b/src/roman_datamodels/_stnode/_schema.py
@@ -526,9 +526,9 @@ class FakeDataBuilder(Builder):
     def make_array(self, schema, defaults):
         import numpy as np
 
-        ndim = _get_keyword(schema, "ndim") or 0
+        ndim = _get_keyword(schema, "ndim") or 1
         dtype = _get_keyword(schema, "datatype") or "float32"
-        shape = [0] * ndim
+        shape = [1] * ndim
         if self._shape is not None:
             for i, v in enumerate(self._shape):
                 if i == len(shape):
--- a/src/roman_datamodels/_stnode/_schema.py
+++ b/src/roman_datamodels/_stnode/_schema.py
@@ -526,6 +526,10 @@ class FakeDataBuilder(Builder):
     def make_array(self, schema, defaults):
         import numpy as np
 
+        # Ideally we default to 0 instead of 1 for ndim and shape here but
+        # cannot until asdf fixes issues with 0-length inline arrays.
+        # Likely we can switch back to 1 when our minimum asdf version is 5.4
+        # which is not yet released.
         ndim = _get_keyword(schema, "ndim") or 1
         dtype = _get_keyword(schema, "datatype") or "float32"
         shape = [1] * ndim
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -22,6 +22,7 @@ import asdf
 import numpy as np
 from asdf.exceptions import ValidationError
 from asdf.tags.core.ndarray import NDArrayType
+from asdf.util import NotSet
 from astropy.time import Time
 
 from roman_datamodels._stnode import NODE_EXTENSIONS, DNode, TaggedObjectNode
@@ -275,13 +276,15 @@ class DataModel(abc.ABC):
         target._files_to_close = []
         target._shape = source._shape
 
-    def save(self, path, dir_path=None, *args, all_array_compression="lz4", **kwargs):
+    def save(self, path, dir_path=None, *args, all_array_compression="lz4", all_array_storage=NotSet, **kwargs):
         path = Path(path(self.meta.filename) if callable(path) else path)
         output_path = Path(dir_path) / path.name if dir_path else path
         ext = path.suffix.decode(sys.getfilesystemencoding()) if isinstance(path.suffix, bytes) else path.suffix
 
         if ext == ".asdf":
-            self.to_asdf(output_path, *args, all_array_compression=all_array_compression, **kwargs)
+            self.to_asdf(
+                output_path, *args, all_array_compression=all_array_compression, all_array_storage=all_array_storage, **kwargs
+            )
         elif ext == ".parquet" and hasattr(self, "to_parquet"):
             self.to_parquet(output_path)
         else:
@@ -297,23 +300,20 @@ class DataModel(abc.ABC):
 
         return asdf.AsdfFile(init, **kwargs)
 
-    def to_asdf(self, init, *args, all_array_compression="lz4", **kwargs):
+    def to_asdf(self, init, *args, all_array_compression="lz4", all_array_storage=NotSet, **kwargs):
         from ._utils import temporary_update_filedate, temporary_update_filename
 
         with (
             temporary_update_filename(self, Path(init).name),
             temporary_update_filedate(self, Time.now()),
         ):
-            all_array_storage = kwargs.pop("all_array_storage", None)
             asdf_file = self.open_asdf(**kwargs)
             asdf_file["roman"] = self._instance
             with asdf.config_context() as cfg:
                 # only set array inline threshold if not already set by the user
-                if cfg.array_inline_threshold is None and all_array_storage is None:
+                if cfg.array_inline_threshold is None and all_array_storage is NotSet:
                     cfg.array_inline_threshold = DEFAULT_ARRAY_INLINE_THRESHOLD
 
-                if all_array_storage is not None:
-                    kwargs["all_array_storage"] = all_array_storage
                 asdf_file.write_to(init, *args, all_array_compression=all_array_compression, **kwargs)
 
     def get_primary_array_name(self):
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -314,7 +314,9 @@ class DataModel(abc.ABC):
                 if cfg.array_inline_threshold is None and all_array_storage is NotSet:
                     cfg.array_inline_threshold = DEFAULT_ARRAY_INLINE_THRESHOLD
 
-                asdf_file.write_to(init, *args, all_array_compression=all_array_compression, **kwargs)
+                asdf_file.write_to(
+                    init, *args, all_array_compression=all_array_compression, all_array_storage=all_array_storage, **kwargs
+                )
 
     def get_primary_array_name(self):
         """
--- a/src/roman_datamodels/datamodels/_core.py
+++ b/src/roman_datamodels/datamodels/_core.py
@@ -34,6 +34,8 @@ __all__ = ["MODEL_REGISTRY", "DataModel"]
 
 MODEL_REGISTRY: dict[str, type[DataModel]] = {}
 
+DEFAULT_ARRAY_INLINE_THRESHOLD = 512
+
 
 def _set_default_asdf(func):
     """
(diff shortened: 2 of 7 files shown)


schlafly (approved):
Thanks Brett!

src/roman_datamodels/_stnode/_schema.py
```diff
@@ -526,9 +526,13 @@ def make_time(self, schema, defaults):
     def make_array(self, schema, defaults):
         import numpy as np
 
-        ndim = _get_keyword(schema, "ndim") or 0
+        # Ideally we default to 0 instead of 1 for ndim and shape here but
+        # cannot until asdf fixes issues with 0-length inline arrays.
+        # Likely we can switch back to 1 when our minimum asdf version is 5.4
+        # which is not yet released.
+        ndim = _get_keyword(schema, "ndim") or 1
```
WilliamJamieson, line 533: Is there any reason to not bump the minimum asdf version?

WilliamJamieson (approved):
Looks good to me.

My only suggestion is to simply bump the minimum ASDF version rather than working around the bug, especially if there is not any usage of Roman in the wild outside of developers and development related activities. In any case the changes are fine.

src/roman_datamodels/datamodels/_core.py
```diff
@@ -298,18 +297,24 @@ def open_asdf(self, init=None, **kwargs):
 
         return asdf.AsdfFile(init, **kwargs)
 
-    def to_asdf(self, init, *args, all_array_compression="lz4", all_array_storage="internal", **kwargs):
+    def to_asdf(self, init, *args, all_array_compression="lz4", **kwargs):
```
WilliamJamieson: Why is the `all_array_storage` option being removed if you are accessing/manipulating it in the function. It should stay in the function's signature as it is a relevant option for the function

src/roman_datamodels/datamodels/_core.py
```diff
@@ -298,18 +297,24 @@ def open_asdf(self, init=None, **kwargs):
...
         from ._utils import temporary_update_filedate, temporary_update_filename
 
         with (
             temporary_update_filename(self, Path(init).name),
             temporary_update_filedate(self, Time.now()),
         ):
+            all_array_storage = kwargs.pop("all_array_storage", None)
```
WilliamJamieson: If this is the case, just set the default for `all_array_storage` to `None` in the function signature.

src/roman_datamodels/datamodels/_core.py
```diff
@@ -298,18 +297,24 @@ def open_asdf(self, init=None, **kwargs):
...
             asdf_file["roman"] = self._instance
-            asdf_file.write_to(
-                init, *args, all_array_compression=all_array_compression, all_array_storage=all_array_storage, **kwargs
-            )
+            with asdf.config_context() as cfg:
+                # only set array inline threshold if not already set by the user
+                if cfg.array_inline_threshold is None and all_array_storage is None:
+                    cfg.array_inline_threshold = DEFAULT_ARRAY_INLINE_THRESHOLD
+
+                if all_array_storage is not None:
```
WilliamJamieson: Why not pass `all_array_storage` into `.write_to`?

src/roman_datamodels/_stnode/_schema.py
braingram: We don't have a version to bump to at the moment. Given the number of issues with 0-sized arrays avoiding using them seems safer.

src/roman_datamodels/datamodels/_core.py
braingram: `all_array_storage` is already covered by `kwargs`. Perhaps we should remove `kwargs` but that's outside the scope of this PR. Setting the default `all_array_storage` to `internal` is problematic for this PR since it overwrites the per-array `inline` storage options.

Maybe it's better if we default this to `NotSet` and keep it in the function signature. I'll take a look.

WilliamJamieson: This discussion is related to asdf-format/asdf#2089.

braingram: I restored `all_array_storage` to the function signature and set it to `NotSet` in https://github.com/spacetelescope/roman_datamodels/pull/686/commits/35dce4d568f522caec04f74b21ba831fa2b751c0

How does that look?

src/roman_datamodels/datamodels/_core.py
```diff
@@ -298,7 +300,7 @@ def open_asdf(self, init=None, **kwargs):
 
         return asdf.AsdfFile(init, **kwargs)
 
-    def to_asdf(self, init, *args, all_array_compression="lz4", all_array_storage="internal", **kwargs):
+    def to_asdf(self, init, *args, all_array_compression="lz4", all_array_storage=NotSet, **kwargs):
```
WilliamJamieson, line 303:
```suggestion
    def to_asdf(self, init, *args, all_array_compression="lz4", all_array_storage=None, **kwargs):
```

src/roman_datamodels/datamodels/_core.py
```diff
@@ -307,9 +309,12 @@ def to_asdf(self, init, *args, all_array_compression="lz4", all_array_storage="i
         ):
             asdf_file = self.open_asdf(**kwargs)
             asdf_file["roman"] = self._instance
-            asdf_file.write_to(
-                init, *args, all_array_compression=all_array_compression, all_array_storage=all_array_storage, **kwargs
-            )
+            with asdf.config_context() as cfg:
+                # only set array inline threshold if not already set by the user
+                if cfg.array_inline_threshold is None and all_array_storage is NotSet:
```
WilliamJamieson, line 314:
```suggestion
                if cfg.array_inline_threshold is None and all_array_storage is None:
```

Note that if we leave things as they are, an end user specifically populating `all_array_storage` with `None` would result in the inline threshold not being set. To me it seems like one would expect passing `all_array_storage=None` to fall back on the defaults instead of turning them off.

src/roman_datamodels/datamodels/_core.py
```diff
@@ -273,12 +276,11 @@ def clone(target, source, deepcopy=False, memo=None):
         target._files_to_close = []
         target._shape = source._shape
 
-    def save(self, path, dir_path=None, *args, all_array_compression="lz4", all_array_storage="internal", **kwargs):
+    def save(self, path, dir_path=None, *args, all_array_compression="lz4", all_array_storage=NotSet, **kwargs):
```
WilliamJamieson, line 279:
```suggestion
    def save(self, path, dir_path=None, *args, all_array_compression="lz4", all_array_storage=None, **kwargs):
```

src/roman_datamodels/datamodels/_core.py
```diff
@@ -22,6 +22,7 @@
 import numpy as np
 from asdf.exceptions import ValidationError
 from asdf.tags.core.ndarray import NDArrayType
+from asdf.util import NotSet
```
WilliamJamieson, line 25:
```suggestion
```

src/roman_datamodels/datamodels/_core.py
```diff
@@ -307,9 +309,12 @@ def to_asdf(self, init, *args, all_array_compression="lz4", all_array_storage="i
...
             asdf_file["roman"] = self._instance
-            asdf_file.write_to(
-                init, *args, all_array_compression=all_array_compression, all_array_storage=all_array_storage, **kwargs
-            )
+            with asdf.config_context() as cfg:
+                # only set array inline threshold if not already set by the user
+                if cfg.array_inline_threshold is None and all_array_storage is NotSet:
+                    cfg.array_inline_threshold = DEFAULT_ARRAY_INLINE_THRESHOLD
+
+                asdf_file.write_to(init, *args, all_array_compression=all_array_compression, **kwargs)
```
WilliamJamieson: Should `all_array_storage` be passed into this call? Otherwise you end up with the `to_asdf` method eating the `all_array_storage` setting.

Note that passing `None` like I have suggested above should not cause issues in ASDF as this option is handled here in `AsdfFile.write_to` https://github.com/asdf-format/asdf/blob/b2d212dc3f7a9df3700a4e13fa51f6aad641cb03/asdf/_asdf.py#L1071-L1072, which calls https://github.com/asdf-format/asdf/blob/b2d212dc3f7a9df3700a4e13fa51f6aad641cb03/asdf/config.py#L361-L364. `None` functions perfectly well in this call and it will not override the default for the config.

WilliamJamieson (commented):
Since `all_array_storge` can really only be one of  the options here: https://github.com/asdf-format/asdf/blob/b2d212dc3f7a9df3700a4e13fa51f6aad641cb03/asdf/typing.py#L67, `NotSet` is overkill as there is no need to distinguish between `None` and the user making no input. Indeed if the user directly passed `None` expecting it to follow normal defaults, the code would misbehave.

braingram:
> Since `all_array_storge` can really only be one of the options here: https://github.com/asdf-format/asdf/blob/b2d212dc3f7a9df3700a4e13fa51f6aad641cb03/asdf/typing.py#L67, `NotSet` is overkill as there is no need to distinguish between `None` and the user making no input. Indeed if the user directly passed `None` expecting it to follow normal defaults, the code would misbehave.

`None` isn't one of the documented possible values:
https://www.asdf-format.org/projects/asdf/en/latest/api/asdf.AsdfFile.html#asdf.AsdfFile.write_to

so we should stick with `NotSet` here.

I updated the test to skip testing `None` which was in part hiding the bug you pointed out where `all_array_storage` was being dropped before getting passed to `write_to` (fixed in 5636d3d).

WilliamJamieson:
> `None` isn't one of the documented possible values: [asdf-format.org/projects/asdf/en/latest/api/asdf.AsdfFile.html#asdf.AsdfFile.write_to](https://www.asdf-format.org/projects/asdf/en/latest/api/asdf.AsdfFile.html#asdf.AsdfFile.write_to)

It is actually documented in the docs: https://www.asdf-format.org/projects/asdf/en/latest/api/asdf.typing.ArrayStorage.html#asdf.typing.ArrayStorage. Via the `ArrayStorage` link in the function signature on the page you linked to. The doc string is not entirely correct.
> 
> so we should stick with `NotSet` here.
> 
> I updated the test to skip testing `None` which was in part hiding the bug you pointed out where `all_array_storage` was being dropped before getting passed to `write_to` (fixed in [5636d3d](https://github.com/spacetelescope/roman_datamodels/commit/5636d3d2a15ca122d0727317cb8e1091f0acff88)).

Not related to this PR. For ASDF I think the discussion in this PR is relevant to this ASDF issue comment https://github.com/asdf-format/asdf/issues/2089#issuecomment-4897556443.

WilliamJamieson (approved):

PR #692, opened 2026-07-17
Handle missing minItems.

schlafly:
This PR intends to improve the handling of read_pattern and other array attributes lacking minItems (c.f. https://github.com/spacetelescope/romancal/pull/2390).  With the existing code in combination with use of create_minimal in the source catalog step to make the segmentation map and source catalog files
https://github.com/spacetelescope/romancal/blob/main/romancal/source_catalog/source_catalog_step.py#L304
https://github.com/spacetelescope/romancal/blob/main/romancal/source_catalog/source_catalog_step.py#L317
array-valued metadata end up getting truncated and lost.

Regression tests are clean: https://github.com/spacetelescope/RegressionTests/actions/runs/30117477542
which points to a separate issue where maybe we should be testing the segmentation maps, and that Brett is working on in adding parquet-based regression tests (c.f. https://github.com/spacetelescope/romancal/pull/2237).

Brett points out that we have a handful of creation routines:
- create_minimal: only minimal required attributes from defaults; previously only took in the minimum number of items, leading to this issue
- create_from_model: copies everything over, including things not in the schema; undesired here because it would bring over many columns that we don't want in the source catalog / segmentation map metadata
- create_fake_data: copies everything over and invents things if necessary to make a valid schema

For this particular application, we want something in between create_minimal and create_from_model, which copies over things present in the rad schema (whether required or not), but not other items absent from the schema.  That is a separate PR.

## Tasks
- [X] Update or add relevant `roman_datamodels` tests.
- [X] Update relevant docstrings and / or `docs/` page.
- [X] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [X] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [X] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

</details

### DIFF
--- a/src/roman_datamodels/_stnode/_schema.py
+++ b/src/roman_datamodels/_stnode/_schema.py
@@ -332,7 +332,10 @@ class Builder:
 
         min_items = _get_keyword(schema, "minItems")
         if min_items is _MISSING_KEYWORD:
-            return arr
+            # The schema imposes no minimum length, so preserve a provided
+            # default (as the scalar builders do) instead of discarding it.
+            # Without a default this yields an empty array.
+            return [copy.deepcopy(sub_default) for sub_default in defaults]
 
         for sub_default in defaults[:min_items]:
             arr.append(copy.deepcopy(sub_default))
--- a/src/roman_datamodels/_stnode/_schema.py
+++ b/src/roman_datamodels/_stnode/_schema.py
@@ -332,10 +332,10 @@ class Builder:
 
         min_items = _get_keyword(schema, "minItems")
         if min_items is _MISSING_KEYWORD:
-            # The schema imposes no minimum length, so preserve a provided
-            # default (as the scalar builders do) instead of discarding it.
-            # Without a default this yields an empty array.
-            return [copy.deepcopy(sub_default) for sub_default in defaults]
+            # No minimum length is imposed by the schema, so keep however many
+            # defaults were provided (as the scalar builders do) instead of
+            # discarding them.  Without a default this yields an empty array.
+            min_items = len(defaults)
 
         for sub_default in defaults[:min_items]:
             arr.append(copy.deepcopy(sub_default))
--- a/src/roman_datamodels/_stnode/_schema.py
+++ b/src/roman_datamodels/_stnode/_schema.py
@@ -335,7 +335,7 @@ class Builder:
             # No minimum length is imposed by the schema, so keep however many
             # defaults were provided (as the scalar builders do) instead of
             # discarding them.  Without a default this yields an empty array.
-            min_items = len(defaults)
+            return copy.deepcopy(defaults)
 
         for sub_default in defaults[:min_items]:
             arr.append(copy.deepcopy(sub_default))
--- a/changes/692.bugfix.rst
+++ b/changes/692.bugfix.rst
@@ -1,0 +1,3 @@
+Preserve provided defaults in ``create_minimal`` for required array attributes
+that lack ``minItems`` in the schema (for example ``exposure.read_pattern`` and
+``cal_logs``) instead of returning an empty array.
--- a/tests/stnode/test_schema.py
+++ b/tests/stnode/test_schema.py
@@ -67,6 +67,10 @@ def test_type(schema, type_):
             {"a": 0},
         ),
         ({"items": {"enum": [0]}}, None, []),
+        # a provided default is preserved when the schema has no minItems
+        ({"items": {"enum": [0]}}, [1, 2], [1, 2]),
+        # nested arrays (e.g. exposure.read_pattern) are preserved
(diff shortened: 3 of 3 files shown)


src/roman_datamodels/_stnode/_schema.py
```diff
@@ -332,7 +332,10 @@ def from_array(self, schema, defaults):
 
         min_items = _get_keyword(schema, "minItems")
         if min_items is _MISSING_KEYWORD:
-            return arr
+            # No minimum length is imposed by the schema, so keep however many
+            # defaults were provided (as the scalar builders do) instead of
+            # discarding them.  Without a default this yields an empty array.
+            min_items = len(defaults)
```
braingram: The general strategy here is that `create_minimal` only uses defaults if the item is required.
read_pattern is required but the items/structure is undefined in the schema:
https://github.com/spacetelescope/rad/blob/f82c94f91e8afabf3348a2be870790c8f129f6b6/latest/meta/exposure.yaml#L261
So the minimal amount that can be taken from the defaults and is valid is an empty read pattern.
I'm ok with the approach in this PR (if the schema has no minItems but a default was provided, use the defaults) given the limited use of lists/arrays in our schemas. An alternative would be to propagate read_pattern in romancal (since I don't see a way for us to define `minItems` or something similar in the schema).
Given the defaults is copied below (341) and then len checked (344) I think we could replace this line with:

```suggestion
            return copy.deepcopy(defaults)
```

schlafly: The test in https://github.com/spacetelescope/romancal/pull/2390 only checks read_pattern but there are a handful of other related cases the code should be able to handle (image_shape; rotation_matrix; dither_executed_pattern).  Do you see a mechanism to propagate these generically in romancal?  It seemed more fragile to list out those cases.

I'm happy with replacing the min_items definition with the direct copy; an earlier version took that approach but I decided it was nicer to follow the path here.  I don't have a strong preference.

Re "The general strategy here is that create_minimal only uses defaults if the item is required."---this change is intended to fix
https://github.com/spacetelescope/romancal/blob/main/romancal/source_catalog/source_catalog_step.py#L317-L319
where conceptually we want to bring over the old metadata into the new object for relevant fields.  In that context, all things being equal, I'd rather also propagate optional fields, and when copying things over, we want to copy all of them.  Do you think that instead we should be using something other than create_minimal there?

braingram: If optional fields are also wanted than maybe this should be using `create_from_model` instead of `create_minimal`?

I tried this PR with a few other attributes and one thing to note is that now `cal_logs` are copied from the defaults for `create_minimal`. I don't think that's an issue (since those should be overwritten when a step finishes).

schlafly: Thanks.  I hadn't thought about create_from_model.  That looks to pull in too much.  For me there are three kinds of fields:
1. required
2. optional --- present in the schema, 'documented,' but not required
3. extra --- things we dropped in but without telling rad about

create_from_model does 1, 2, and 3; create_minimal does 1 (with caveats about things like read_pattern where minItems comes into play).  There aren't currently any fields in class 2 that are relevant for source catalog / segmentation map construction.  There are fields in class 3 (e.g., wcs_fit_results), however, so I am leaning toward create_minimal + this PR rather than create_from_model.

Is there a concrete case where the 'take the minimum number of items only' behavior is helping us?  I'm trying to think of a case where the default is incompatible with the min / max items and we want it cut down and I'm not immediately coming up with one.

braingram: > Is there a concrete case where the 'take the minimum number of items only' behavior is helping us? I'm trying to think of a case where the default is incompatible with the min / max items and we want it cut down and I'm not immediately coming up with one.

Not that I'm aware of. I don't have any objections to changing this (as is done in this PR). It does mean we handle arrays differently than objects (where we ignore provided defaults for non-required items) but that seems fine since they're used differently within rdm. Looking at the various `type: array` entries we have in schemas that don't have `minItems` they all look like things we'd want to propagate if provided as a default.

More generally, I wonder if it would be better to have an API that's something like:

- Model.create_minimal: no defaults, only schema/model-defined items
- Model.create_from_model: defaults/model and some strategy "required", "optional", "all"
- Model.create_fake_data: optional model, perhaps only with strategy "all"

but that is beyond the scope of this PR.

PaulHuwe: I am in favor of exploring the IDE expansion you suggest, @braingram

PaulHuwe (approved):
LGTM

